PR #11, opened 2019-10-30
Minor modifications for task EXTENSION-3513

adelego:
I've updated the baseButton style to match Figma's [design](https://www.figma.com/file/HXSuovuznNj7bMSQCu9oOB/Web-Platform-Design-System?node-id=931%3A526)
Screenshot 2019-10-30 at 12 03 16


src/atoms/svg/icons/Icons.mdx
```diff
@@ -42,4 +43,6 @@ Icons can be configured in the same way (have the same props) as the SvgIcon com
   <NoNetwork size={2} />
   <Shared size={2} />
   <Loader size={2} />
+  <BigLogo size={1} />
```
maryiam: What is BigLogo ?

src/ions/button/base/BaseButton.tsx
```diff
@@ -53,15 +54,15 @@ const Button = styled.button<BaseProps>`
     color: ${getHoverColor};
     border-color: ${getHoverBorderColor};
     background-color: ${getHoverBackground};
-    transition: border-color .1s ease-in-out, background-color .1s ease-in-out;;
+    transition: border-color 0.1s ease-in-out, background-color 0.1s ease-in-out;
```
maryiam: Do you have some local style-lint ?

src/atoms/svg/icons/Icons.mdx
adelego: My bad, I wanted to add the [Dashlane logo](https://www.figma.com/file/HXSuovuznNj7bMSQCu9oOB/Web-Platform-Design-System?node-id=931%3A485) but it doesn't fit this component as it isn't squared. I forgot to remove it completely from the docz

src/ions/button/base/BaseButton.tsx
adelego: It looks like it

maryiam: Can you add it to the lib in this case? It would be nice that we all follow the same pattern and the one you are using seems to be a good one.

PR #15, opened 2019-11-25
[WIP] feat(shared-design-system): create stylesheet generator

adelego:
I've generated css and scss variables stylesheets with plop
TODO:

- [x] Expose these stylesheets in package
- [ ] Write documentation on how to use them in the project
- [ ] Generate on publish?

### DIFF
--- a/config/generator/colorGenerator/generateColorStylesheets.js
+++ b/config/generator/colorGenerator/generateColorStylesheets.js
@@ -13,7 +13,9 @@ const generateStylesheets = async () => {
     await Promise.all(
       languages.map(async language => generateStylesheet(language))
     );
-    console.log('All done ✅ \nAll Icon components have been generated 🙌🏼.');
+    console.log(
+      'All done ✅ \nAll colors have bee added to your stylesheets 🙌🏼.'
+    );
   } catch (e) {
     console.error(`Process terminated with error ${e}`);
   }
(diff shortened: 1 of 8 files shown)


config/generator/colorGenerator/generateColorStylesheets.js
```diff
@@ -0,0 +1,22 @@
...
+};
+
+const languages = ['css', 'scss'];
+
+const generateStylesheets = async () => {
+  try {
+    await Promise.all(
+      languages.map(async language => generateStylesheet(language))
+    );
+    console.log('All done ✅ \nAll Icon components have been generated 🙌🏼.');
```
maryiam: Typo: replace the sentence with the message that we want to display after generating the colors

plopfile.js
```diff
@@ -28,4 +30,26 @@ module.exports = plop => {
...
+        name: 'language'
+      }
+    ],
+    actions: [
+      {
+        type: 'add',
+        data: {
+          colorTokens
+        },
+        path: './colors.{{language}}',
```
maryiam: Maybe put these files in the colors folder (I think we have already one under src) ?

config/generator/colorGenerator/templates/colors.css.hbs
```diff
@@ -0,0 +1,10 @@
+:root {
```
maryiam: I think that theses files need to be added to the published ones (as we previously discussed). So maybe add them to the section `files` in `package.json`.

maryiam: Also, concerning when to generate these files. I think that it would be enough, for now, to simply generate them and add them to git (so they will be added in the next published version). We will see later how to automatize this on the CI.

PR #16, opened 2019-12-02
fix(yarn): remove usages of Dashlane internal registry

ncordin:


maryiam:
Would you please retry with signed commits? Thanks

PR #18, opened 2019-12-16
Add check icon

adelego:
We need to add a new SVG icon for this task: https://jira.dashlane.com/browse/EXTENSION-3518


src/atoms/svg/icons/Icons.mdx
```diff
@@ -36,12 +37,13 @@ Icons can be configured in the same way (have the same props) as the SvgIcon com
...
-  <Edit size={2} title="translation_key" />
+  <Edit size={2} title='translation_key' />
   <CopyPassword size={2} />
   <Hide size={2} />
   <Reveal size={2} />
   <NoNetwork size={2} />
   <Shared size={2} />
   <Loader size={2} />
   <CaretDown size={2} />
+  <Check size={2} style={{ fill: 'none' }} />
```
b-jan: Avoid using extra unneeded prop here

adelego: As discussed, this should be in the generated svg path. 
I will open an other PR to add a property fill in the icon-def.json and generate the icons with this fill property if it's present

PR #21, opened 2020-02-07
Fix color export

adelego:
With the previous settings, we are exporting a typescript file whereas it should be compiled to js. I've added it to the export file


package.json
```diff
@@ -8,7 +8,7 @@
   "files": [
     "lib",
     "CONTRIBUTING.md",
-    "src/colors/colors.*",
+    "src/colors/colors.(?!ts).*$",
```
b-jan: why do you change this one?

adelego:
As discussed with @maryiam, it's better to have two entry points on the package:
- one for components
- one for design-tokens like colors and later on font style, spacing...

It's also better that projects installing `ui-components` import transpiled files.

What we should do in a next pull-request is:
 - create a JS file in `design-tokens/colors`
- expose color files from '@dashlane/ui-components/design-tokens`

PR #22, opened 2020-02-13
feat: add open website icon

adelego:
We need a new icon for [ticket](https://jira.dashlane.com/browse/EXTENSION-4090)


src/atoms/icons/OpenWebsiteIcon.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import SvgWrapper from '@ions/icon/wrapper/SvgWrapper';
+
+const OpenWebsiteIcon: React.FunctionComponent<
+  import('../../ions/icon/types').IconProps
```
Domino9697: Is there a reason we are doing the import here ?

maryiam: The relative imports are not accepted by ESlint and then if we do an import with an alias (like @ions/...) then this import will remain the same in the generated type file OpenWebsiteIcon.d.ts (typescript doesn’t process these alias before generating the declaration type files)

src/design-tokens/icons/icons-defs.json
```diff
@@ -1,39 +1,40 @@
...
+	"edit": "<path fillRule=\"evenodd\" clipRule=\"evenodd\" d=\"M14.8833 1L19 5.11667L16.6355 7.48114L15.5802 6.42584L16.8894 5.11667L14.8833 3.1106L13.5742 4.41977L12.5189 3.36447L14.8833 1ZM11.2097 4.67365L15.3264 8.79032L6.48889 17.6278L1 19L2.37222 13.5111L11.2097 4.67365ZM3.71982 14.2741L3.05113 16.9489L5.7259 16.2802L13.2158 8.79032L11.2097 6.78424L3.71982 14.2741Z\"/>",
```
maryiam: Why do we have a different formatting on this file now

PR #23, opened 2020-02-18
Dropdown menu

adelego:
This is a PR to add dropdown menus to the lib.
[Designs link](https://www.figma.com/file/I22yz2CF3Tnn2yr13snOsi/Web-UI-Kit?node-id=988%3A0)
Screenshot:
Capture d’écran 2020-02-18 à 12 10 49
Capture d’écran 2020-02-19 à 14 12 53

DONE:
- [x] Create Dropdown Menu atom and necessary ions
- [x] Create Quick Actions Menu
- [x] Handle positionning
- [x] Make component usability more flexible
- [x] Write a complete documentation

TODO:
- [ ] Write tests


src/atoms/dropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,141 @@
...
+    e.preventDefault();
+    setDropdownIsOpen(!dropdownIsOpen);
+  };
+
+  const getChild = () => {
+    let child;
+    try {
+      React.Children.only(props.children);
+
+      child = React.Children.map(props.children, child => {
```
Aktanee: Why not return directly without using child ?

src/atoms/dropdownMenu/helpers.ts
```diff
@@ -0,0 +1,100 @@
...
+): Placement | null => {
+  if (!placement) {
+    throw new Error(
+      'Dropdown cannot be displayed in the container Element/Window due to its incompatible size.' +
+        'Please consider reducing it or better position tooltiped' +
+        'element in the screen.'
+    );
+  }
+
+  if (null === menuRef.current) {
```
Aktanee: Do we need to throw an error if there is no menuRef si we don't return anything ?

src/utility/withClickOutside.tsx
```diff
@@ -0,0 +1,50 @@
...
+                </div>
+            ) : (
+                <button onClick={() => setOpen(true)}>Open </button>
+            )}
+            </div>
+        );
+    }
+*/
+
+export const useOnClickOutside = (
```
Aktanee: :star-struck:

src/atoms/dropdownMenu/DropdownMenu.mdx
```diff
@@ -0,0 +1,228 @@
+---
+name: Dropdown Menu
+menu: Components
+---
+
+import { Playground, Props } from 'docz';
+import DropdownMenu from './DropdownMenu';
+import DropdownElement from '../../ions/dropdown/element/DropdownElement';
```
b-jan: Could we have absolute path instead?

src/atoms/dropdownMenu/DropdownMenu.mdx
```diff
@@ -0,0 +1,228 @@
...
+
+## Properties
+
+<Props of={DropdownMenu} />
+
+## Usage
+
+The dropdown can be defined alongside a controller. To display it correctly:
+
+- **both components must be wrapped in a container with an relative or absolute position**
```
b-jan: typo: `with A relative`

src/atoms/dropdownMenu/DropdownMenu.mdx
```diff
@@ -0,0 +1,228 @@
...
+  }}
+>
+  {() => {
+  const [dropdownIsOpen, setDropdownIsOpen] = React.useState(false);
+    const controllerRef = React.useRef(null);
+
+return (
+
+<div style={{ position: 'relative' }}>
+<div ref={controllerRef}>
```
b-jan: Why don't we have indentation here?

src/atoms/dropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+      menuRef,
+      props.controllerRef
+    );
+
+    const containerElement =
+      (props.containerRef && props.containerRef.current) || window;
+    const containerDimensions = getContainerBoundaries(containerElement);
+
+    try {
+      const newPlacement = getElementPlacement(
```
Domino9697: I would avoid having so many arguments for a single function. It makes it difficult to understand without looking at its signature.

src/atoms/dropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+          <DropdownCard maxHeight={props.menuMaxHeight}>
+            {props.children}
+          </DropdownCard>
+        </MenuWrapper>
+      )}
+    </>
+  );
+};
+
+export default DropdownMenu;
```
Domino9697: I think we should memoize this component. I am thinking about the Popup. We will have a dropdown for every credential row of our list.

src/atoms/dropdownMenu/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+          elementDimensions.totalHeight >
+        containerDimensions.maxHeight
+      );
+    default:
+      assertUnreachable(placement);
+      return true;
+  }
+};
+
+export const getControllerAndMenuDimensions = (
```
Domino9697: Again here I would transform the arguments in a single object for more clarity

src/atoms/dropdownMenu/helpers.ts
```diff
@@ -0,0 +1,67 @@
...
+  menuRef: React.RefObject<HTMLElement>,
+  buttonRef: React.RefObject<HTMLElement>,
+  elementIsBefore?: boolean
+): ElementDimensions => {
+  const dropdownStyle = getCSSComputedStyle(menuRef.current, elementIsBefore);
```
Domino9697: Here I expect the `getCSSComputedStyle` to take the ref as the first argument so this is good. But I think the elementIsBefore property should be in an options object. What do you think ?

src/atoms/tooltip/Tooltip.tsx
```diff
@@ -150,72 +148,48 @@ class Tooltip extends React.PureComponent<Props, State> {
...
-  private positionTooltip() {
+    const containerElement =
+      (this.props.options &&
+        this.props.options.container &&
+        this.props.options.container.current) ||
+      window;
+    const containerDimensions = getContainerBoundaries(containerElement);
     try {
-      const placement = this.getTooltipPlacement(
+      const placement = getElementPlacement<Placement>(
```
Domino9697: Why do we need to specify the `Placement` type ?

src/atoms/tooltip/Tooltip.tsx
```diff
@@ -150,72 +148,48 @@ class Tooltip extends React.PureComponent<Props, State> {
...
     } catch (e) {
-      console.error(e);
+      console.error('Error while computing DropdownMenu position', e);
     }
   }
 
-  private cloneChildren = (title: string) => (child: React.ReactNode) => {
+  private cloneChildren = (title: string) => (
+    child: React.ReactNode,
+    key: number
```
Domino9697: Maybe we can rename this to `childIndex` ?

src/atoms/tooltip/helpers.ts
```diff
@@ -117,156 +118,65 @@ export const getTooltipBgColorByType = (type: TooltipType) => {
...
-  placement: Placement;
-}): boolean => {
-  let { totalWidth, totalHeight } = measurements;
-  let { maxHeight, minHeight, maxWidth, minWidth } = getContainerBoundaries(
-    container
-  );
-
+export const isTooltipOutOfContainer: IsElementOutOfContainerMethod<
+  Placement
+> = ({ elementDimensions, containerDimensions, placement }): boolean => {
```
Domino9697: Since we already type the `const`, I don't think we need the return type `boolean`

src/molecules/quick-actions-menu/QuickActionsMenu.mdx
```diff
@@ -0,0 +1,37 @@
...
+
+import { Playground, Props } from 'docz';
+import Button from '../../atoms/button/Button';
+import MoreIcon from '../../atoms/icons/MoreIcon';
+import DropdownElement from '../../ions/dropdown/element/DropdownElement.tsx';
+import QuickActionsMenu from './QuickActionsMenu';
+
+# Quick Actions Menu
+
+Quick action menu is a specific components meant for users to quickly copy
```
Domino9697:
```suggestion
Quick action menu is a specific component meant for users to quickly copy
```

src/molecules/quick-actions-menu/QuickActionsMenu.tsx
```diff
@@ -0,0 +1,52 @@
...
+  width: 32px;
+  height: 32px;
+  position: relative;
+`;
+
+const QuickActionsMenu: React.FC<Props> = props => {
+  const [isOpen, setIsOpen] = React.useState(false);
+  const controllerRef = React.useRef<HTMLDivElement>(null);
+
+  const onButtonClick = (e: React.MouseEvent<HTMLButtonElement>) => {
```
Domino9697: We can use the `useCallback` hook here

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
+/** Both tooltips and dropdown use the DOM to make sure
+ * they are fitting in a container
+ */
+
+const containsNumber = (value: string) => /\d/.test(value);
+const containsChar = (value: string, char: string) =>
+  value.indexOf(char) !== -1;
+
+export const checkIsStyleComputed = (style: CSSStyleDeclaration) => {
```
Domino9697: Maybe we should rename this to `isHeightAndWidthSet`

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  return returnNumber(width);
+};
+
+export const getAlternativeStyle = (
+  initialStyle: CSSStyleDeclaration,
+  element: EventTarget | null
+) => {
+  let innerDiv = document.createElement('div');
+  (Object.values(initialStyle) as (keyof CSSStyleDeclaration)[]).forEach(
```
Domino9697: This feels weird ... Is TypeScript not able to infer this type by itself ?

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+        property &&
+        property !== 'length' &&
+        property !== 'parentRule' &&
+        !containsNumber(property.toString())
+      ) {
+        innerDiv.style[property] = initialStyle[property];
+      }
+    }
+  );
+  (element as Element).appendChild(innerDiv);
```
Domino9697: I do not understand. Element can be `null` apparently. Why can we safely assert that this is an `Element` ?

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+): ContainerDimensions => {
+  if (container === window) {
+    return {
+      maxHeight: window.innerHeight,
+      minHeight: 0,
+      maxWidth: window.innerWidth,
+      minWidth: 0
+    };
+  }
+  const containerRect = getBoundingRect(container as Element);
```
Domino9697: I don't like these type assertions :) I believe here we know that the container is an Element and not the window !
If that is the case, we can do a type guard 

```suggestion
  const isWindow = (element: Element | Window): element is Window => container === window;

  if (container === window) {
    return {
      maxHeight: window.innerHeight,
      minHeight: 0,
      maxWidth: window.innerWidth,
      minWidth: 0
    };
  }
  const containerRect = getBoundingRect(container as Element);
```

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+    maxHeight: containerRect.top + containerRect.height,
+    minHeight: containerRect.top,
+    minWidth: containerRect.left,
+    maxWidth: containerRect.left + containerRect.width
+  };
+};
+
+export const getCSSComputedStyle = (
+  ref: Element | null,
+  getBeforePseudoElement?: boolean
```
Domino9697: I must have said this elsewhere but when I read the function name, I expect it to receive a single argument (the reference). Anything else is not clear and should be considered as options I believe. 
I would therefore put an `options` object as second argument that includes the `getBeforePseudoElement`

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+) =>
+  window.getComputedStyle(
+    ref as Element,
```
Domino9697: Does that work even when ref is null ?

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+    getBeforePseudoElement ? ':before' : undefined
+  );
+
+export const getElementDimensions = (
+  menuRef: React.RefObject<HTMLElement>,
+  elementIsBefore?: boolean
```
Domino9697: Same comment here ;)

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+}
+
+interface ContainerDimensions {
+  maxHeight: number;
+  maxWidth: number;
+  minHeight: number;
+  minWidth: number;
+}
+
+export type IsElementOutOfContainerMethod<T> = (params: {
```
Domino9697: Nice use of the generic type ! 

Now I believe, there are two types of `Placement` am I wrong ? Maybe we could have a union type of all of these and then say that this generic type extends the union type  🙂

src/utility/positionCompute.ts
```diff
@@ -0,0 +1,197 @@
...
+  elementDimensions: ElementDimensions;
+  containerDimensions: ContainerDimensions;
+  placement: T;
+}) => boolean;
+
+export const getElementPlacement = <T>(
```
Domino9697: Same comment here

src/atoms/tooltip/Tooltip.tsx
adelego: Because the Placements options are not the same for the Tooltips and the Dropdowns

adelego: We have top, bottom, left and right in one case, just top and bottom in the other

src/atoms/dropdownMenu/DropdownMenu.mdx
adelego: Prettier doesn't like functions inside Playgrounds :/

src/atoms/dropdownMenu/DropdownMenu.tsx
adelego: You are right, I'm changing it

src/atoms/dropdownMenu/DropdownMenu.tsx
adelego: I don't think it will work properly as we pass children as props as well

adelego: As discussed I'm not going to use React.memo because DropdownMenu has children and I'm afraid it would be counter-productive to compare children on every render.

src/atoms/tooltip/Tooltip.tsx
adelego: Force of habit to name this parameter `key` ;)

src/utility/positionCompute.ts
adelego: No, the type is inferred as a string

src/utility/positionCompute.ts
adelego: I've added some guard safe here 👍

src/utility/withClickOutside.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
```
maryiam: If this file is not used anymore we can remove it

src/atoms/tooltip/Tooltip.tsx
```diff
@@ -153,73 +151,49 @@ class Tooltip extends React.PureComponent<Props, State> {
...
+        placement: this.props.placement || this.defaultPlacement,
+        possibilities: this.tooltipPossiblePlacements,
+        elementDimensions,
+        containerDimensions
+      });
       this.setState({ placement: placement });
     } catch (e) {
       this.setState({ outOfContainer: true });
-      console.error(e);
+      console.error('Error while computing DropdownMenu position', e);
```
maryiam: The message should be related to Tooltip and not DropdownMenu

src/atoms/dropdownMenu/DropdownMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+import DropdownCard from '@ions/dropdown/card/DropdownCard';
+import DropdownElement from '@ions/dropdown/element/DropdownElement';
+import styled from 'styled-components';
+import { DropdownElementInfo, OnClickFunction } from './types';
+import { useOnClickOutside } from '@utility/withClickOutside';
+
+interface Props {
+  elements: DropdownElementInfo[];
+  menuMaxHeight?: number;
+  dockingSide?: 'left' | 'right';
```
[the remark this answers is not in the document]
adelego: It is already 🤔

PR #25, opened 2020-03-04
fix: tooltip opacity and out of container display

maryiam:
- Fix tooltip opacity bug when the button is disabled: wrapped the button child with a `div` 
- Fix tooltip being displayed out of container: the placement is correctly calculated and the `Tooltip` component throws when there is no placement possible in the current container but the styled component wrapper TooltipWrapper kept displaying the tooltip based on the user given placement. Fixed this by making styled-component `TooltipWrapper` component aware when there is no possible placement. So now, when there is no possibility to display the tooltip in the container, we simply keep the tooltip hidden and the error will be displayed in the console

### DIFF
--- a/src/atoms/tooltip/Tooltip.tsx
+++ b/src/atoms/tooltip/Tooltip.tsx
@@ -6,6 +6,7 @@ import {
   checkIsStyleComputed,
   computeTooltipHeight,
   computeTooltipWidth,
+  displayIfInContainer,
   getAlternativeStyle,
   getBoundingRect,
   getCSSComputedStyle,
@@ -46,9 +47,12 @@ export interface Props {
 
 interface State {
   placement: Placement;
+  outOfContainer: boolean;
 }
 
-export type WrapperProps = Omit<Required<Props>, 'title'>;
+export type WrapperProps = Omit<Required<Props>, 'title'> & {
+  outOfContainer: boolean;
+};
 
 const TooltipWrapper = styled.span<WrapperProps>`
   > :first-child:hover {
@@ -85,8 +89,7 @@ const TooltipWrapper = styled.span<WrapperProps>`
   }
 
   > :first-child:hover::before {
-    visibility: visible;
-    opacity: 1;
+    ${displayIfInContainer}
   }
 
   > :first-child::after {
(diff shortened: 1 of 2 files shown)


src/atoms/tooltip/Tooltip.tsx
```diff
@@ -228,16 +232,12 @@ class Tooltip extends React.PureComponent<Props, State> {
       options = {}
     } = this.props;
     const childrenWithProps = React.Children.map(
-      children,
+      <div>{children}</div>,
```
Domino9697: I think we need to explain the intent of the div, otherwise a developer will just remove it at some point. Now, I would not put a comment here to say that we need a div (let's try and avoid comments as much as possible)
Maybe we could add a class here in the div to set the opacity to 1. What do you think ?

maryiam:
> It's a bit weird to have `displayIfInContainer` and `outOfContainer` why not use `insideContainer` ?

I named it `outOfContainer` because it's aligned with the logic of catching the error regarding the placement and set the tooltip state as out of the container. It would be weird to set there `insideContainer` false.

src/atoms/tooltip/Tooltip.tsx
```diff
@@ -228,16 +232,12 @@ class Tooltip extends React.PureComponent<Props, State> {
...
       this.cloneChildren(title)
     );
 
     return (
-      <TooltipWrapper
-        type={type}
-        placement={this.state.placement}
-        options={options}
-      >
+      <TooltipWrapper type={type} {...this.state} options={options}>
```
Domino9697: I don't understand why we removed the placement here

src/atoms/tooltip/Tooltip.tsx
maryiam: Why do you mean by `class`? we are not working with `css` here. You mean wrap it with a styled Component rather than a `div`? I don't get the idea of setting the opacity as 1 for the div.

Domino9697: Yes you are right ... I was just wondering if we could put something in this div to avoid having it like this. It is tempting to put a fragment in its stead ...
I guess a comment would be better suited

maryiam: I hate comments too. But after thinking I will not add unneeded styled-component to replace the div because it will be useless. I'll add a comment.

PR #28, opened 2020-03-24
Feature/textinput css

pducolin:
Adding TextInput based on ReactJS + TypeScript + Sass


src/components/TextInput/TextInput.scss
```diff
@@ -0,0 +1,85 @@
...
+  display: flex;
+  align-items: center;
+
+  /* border */
+  border: 1px solid colors.$dash-green-03;
+  border-radius: 4px;
+  box-sizing: border-box;
+}
+
+.input:focus {
```
oclero: You can enjoy the power of Sass nesting feature:
```scss
.input {
  ...

  &:focus {
    /*  code for focus pseudo-state */
    ...
  }

  /*  same for other pseudo-states */
  ...
}
```

src/components/TextInput/TextInput.scss
```diff
@@ -0,0 +1,85 @@
...
+  border: 1px solid colors.$mid-green-00;
```
oclero: If the border's width never changes, you can set it once for all in the base CSS, then only change the color with `border-width`. FYI, `border` is just an alias for 3 properties: `border-width`, `border-style`, `border-color`.

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const TextInput: FC<TextInputProps> = props => {
+  const className = getClassNameFromFeedbackType(props.feedbackType);
+
+  return (
+    <div className={'root ' + className}>
+      <input {...props} className={'input ' + className} type='text' />
+      {shouldShowFeedback(props) ? (
+        <span className={'feedback ' + className}>{props.feedbackText}</span>
+      ) : null}
+    </div>
```
oclero: You can use my utility function to simplify code here.

pducolin: It is added in the followinf PR

src/components/TextInput/TextInput.mdx
```diff
@@ -0,0 +1,31 @@
+---
+name: TextInput
+menu: Components
+---
+
+import { Playground, Props } from 'docz';
+import { TextInput } from '../TextInput';
```
sericaia: replace by `import { TextInput } from './TextInput';`

pducolin: good catch!

src/components/TextInput/TextInput.scss
```diff
@@ -0,0 +1,81 @@
+@use '../../colors/colors';
```
b-jan: I suggest to import your mixins globally, to avoid repeating those lines.

docs/assets.json
```diff
@@ -1,21 +1,23 @@
 {
```
goldins: Should this file be ignored? The directory is in `gitignore`

package.json
```diff
@@ -26,10 +26,12 @@
     "check-lint": "eslint --ext tsx,ts,js,json .",
     "exec-docz-update": "node config/updateDoczBuild.js",
     "generate-icon-components": "node config/generator/icons/generateIcons.js",
-    "generate-colors-stylesheets": "node config/generator/colors/generateColorStylesheets.js"
+    "generate-colors-stylesheets": "node config/generator/colors/generateColorStylesheets.js",
+    "sass:watch": "yarn sass --watch src:src"
   },
   "sideEffects": false,
   "dependencies": {
+    "@babel/runtime": "^7.9.2",
```
maryiam: why do we need `@babel` now?  The project used to depend on TS transpilation only.

src/typography/_typography.scss
```diff
@@ -0,0 +1,15 @@
+@mixin body-medium {
+  font-family: 'Public Sans';
```
goldins: May want to share the common styling across `body-small` and `body-medium`

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+  error: 'error',
+  warning: 'warning'
+};
+
+const getClassNameFromFeedbackType = (feedbackType?: TextInputFeedbackType) => {
+  return feedbackType ? classNameFromFeedbackType[feedbackType] : '';
+};
+
+const shouldShowFeedback = ({ feedbackText, feedbackType }: TextInputProps) => {
+  return feedbackText && feedbackType;
```
goldins: Not really a functional bug but this would return a string which might cause issues later. I suggest casting the return value (and potentially defining it in TS as well)

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import './TextInput.css';
+
+type TextInputFeedbackType = 'error' | 'warning';
+
+type _ForbiddenProps = 'size' | 'prefix' | 'type';
```
goldins: Why are these forbidden?

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import './TextInput.css';
```
goldins: I _think_ there's a way to import SASS files which would improve the developer experience. Can you look into that?

goldins: You might just need sass-loader: https://webpack.js.org/loaders/sass-loader/

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+type _ForbiddenProps = 'size' | 'prefix' | 'type';
+
+interface TextInputProps
+  extends Omit<React.InputHTMLAttributes<HTMLInputElement>, _ForbiddenProps> {
+  placeholder?: string;
+  feedbackType?: TextInputFeedbackType;
+  feedbackText?: string;
+}
+
+const classNameFromFeedbackType: Record<TextInputFeedbackType, string> = {
```
maryiam: If the names are exactly the same why not use directly `props.feedbackType` as a className?

tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "jsx": "react",
     "moduleResolution": "node",
     "strictNullChecks": true,
+    "allowSyntheticDefaultImports": true,
```
maryiam: Why do we need this?

src/components/TextInput/TextInput.scss
goldins: ^ I disagree because that might make it harder to track variables down to the source, especially if we end up with duplicated names; I'd prefer to be explicit.

Though I haven't used SASS in a while, supposedly they have namespaces now?

src/components/TextInput/TextInput.tsx
goldins: I think if you add `sass-loader` you can import sass files directly, which would improve the developer experience.

maryiam:
Is the build process adapted now to include css with the components?

src/colors/colors.css
```diff
@@ -1,72 +0,0 @@
-:root {
```
maryiam: This file is actually generated. It will appear again if someone runs the command `yarn generate-colors-stylesheets`

src/components/TextInput/TextInput.tsx
maryiam: Yes! and we can configure absolute paths with sass-loader (in the @use syntax as well I believe).

src/colors/colors.css
pducolin: ok, I guess we still need it until we migrate away from s-c, will amend this

tsconfig.json
pducolin: to have ES6 style imports 
`import React from 'react'`
But as jest is not yet compatible I'll amend this

src/typography/_typography.scss
pducolin: I prefer duplication here (aha programing).
Today they share font-family, but things can change in the future

docs/assets.json
pducolin: removed

pducolin:
This PR is blocked by rollup configuration not working with CSS/SCSS

package.json
pducolin: Without it I get the following error at runtime on docz:
```./src/atoms/tooltip/Tooltip.mdx
Module not found: Can't resolve '/Users/pducolin/repos/ui-components/node_modules/babel-preset-react-app/node_modules/@babel/runtime/helpers/esm/defineProperty' in '/Users/pducolin/repos/ui-components/src/atoms/tooltip'```

Comment (pducolin) [reply package.json]: `Tooltip` is in the styled-component library, so it's not related to what we added

Comment (pducolin) [reply package.json]: Maybe it's the TS step I'm missing

File: src/typography/_typography.scss (hunk 8 above)
Comment (oclero) [reply src/typography/_typography.scss]: I agree with @pducolin . Factorizing everything just for the sake of factorizing will prevent flexibility. :+1:

File: package.json (hunk 7 above)
Comment (pducolin) [reply package.json]: It's not TS. I can't have docz building without this dependency 🤷‍♀️

**Comment by pducolin** [comment]
> Some Scss code can be improved, and the rest is good.

File: doczrc.js [hunk 15]
```diff
@@ -12,6 +12,13 @@ export default {
   hashRouter: true,
   base: '/ui-components/',
   modifyBundlerConfig: config => {
+    const rules = [
```
maryiam: It's better to use [docz-plugin-css](https://github.com/doczjs/docz-plugin-css) here

maryiam: you can configure it to understand `sass` syntax as well.

src/components/TextInput/TextInput.tsx
pducolin: we don't want those properties from HTML `input` to be accessible by the dev using the component. We don't want a fixed size, nor a prefix, nor to change the type - fixed to text.

src/colors/colors.css
ni0ki: Can confirm, the popup postcss directly references it, so it should be left untouched

src/components/TextInput/TextInput.module.scss
```diff
@@ -0,0 +1,81 @@
...
+    margin-top: 4px;
+  }
+
+  .feedback.warning {
+    color: colors.$orange-00;
+  }
+
+  .feedback.error {
+    color: colors.$functional-red-02;
+  }
```
oclero, line 80: You can nest these ones into `.feedback` like this:
```scss
 .feedback {
    @include typography.body-small;
    color: colors.$black;
    margin-top: 4px;

    &.warning {
      color: colors.$orange-00;
    }

    &.error {
      color: colors.$functional-red-02;
    }
  }
```

src/components/index.ts
```diff
@@ -1,4 +1,5 @@
 export { default as DropdownMenu } from './dropdown/DropdownMenu';
 export { default as DropdownElement } from './dropdown/DropdownElement';
 export { default as Tooltip } from './tooltip/Tooltip';
+export { TextInput } from './textinput/TextInput';
```
sericaia, line 4: should be `./TextInput/TextInput`

pducolin: I changed to `textinput` to be consistent with already existing components. But I'm happy to change them all to CamelCase

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import * as style from './TextInput.module.scss';
```
sericaia, line 2: I am wondering if we can use styles instead of style since is what we use in leeloo

src/components/index.ts
sericaia: ah I see. Maybe you should do text-input then

pducolin: You're right, just noticed `quick-action-menu` naming. Changing it

package.json
pducolin: For documentation, solution was `rm -rf .docz && rm -rf node_modules && yarn`

src/colors/colors.css
pducolin: Restored!

src/components/TextInput/TextInput.tsx
pducolin: Changed

src/components/TextInput/TextInput.module.scss
pducolin: For documentation, sass nesting doesn't work well with css modules

src/typography/_typography.scss
adelego: Complete font-family is `'Public Sans', 'Helvetica Neue', Helvetica, Arial, 'Lucida Grande', sans-serif`
I don't mind duplication though

tsconfig.json
adelego: FYI we found a fix for this bug in jest yesterday: https://forums.expo.io/t/jest-tsx-expo-34-with-typescript-renderer-create-error-invariant-violation-element-type-is-invalid-expected-a-string-for-built-in-components-or-a-class-function-for-composite-components-but-got-object/26787/2

src/components/TextInput/TextInput.scss
adelego: Popup versus Leeloo practise here 😉 
We're too lazy to import our variables in the Popup in every css file!
But i agree with you Simon that we don't know what happens when two variables have the same name and we don't have auto-complete either :/

src/components/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const shouldShowFeedback = ({ feedbackText, feedbackType }: TextInputProps) => {
+  return feedbackText && feedbackType;
+};
+
+export const TextInput: FC<TextInputProps> = props => {
+  const className = getClassNameFromFeedbackType(props.feedbackType);
+
+  return (
+    <div className={'root ' + className}>
```
adelego: Can we use `root ${className}` here?
If you always need to prepend an attribute to classNames, it would be nice to have a method to do it rather than joining strings in components

src/components/text-input/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export const TextInput: React.FC<TextInputProps> = props => {
+  const styledClassName =
+    props.feedbackType && styles[props.feedbackType]
+      ? styles[props.feedbackType]
+      : '';
+
+  return (
+    <div className={`${styles.root} ${styledClassName}`}>
```
sericaia, line 26: maybe you can use  `<div className={getClassNames(styles.root, styles.somethingElse)}>`

src/utility/cssUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Generates the CSS class names, separated by a space character.
+ * @param classes A CSS class (`string`), or an `Object` with CSS classes as keys,
+ * and expressions evaluable to `boolean` as values.
+ * If the expression is evaluated to `true`, the CSS class will be added.
+ *
+ * Example: `getClassNames("class-1", { "class-2": true, "class-3": false});`
+ */
+export const getClassNames = (...classes: (string | CssClasses)[]): string => {
```
adelego, line 13: We use t[his package](https://github.com/JedWatson/classnames#readme) in the Popup. Why did you feel the need to do it yourself?

oclero: 'It's my code) IMO, the less dependencies we have, the better. We don't need to pull one more dependency for just one basic function that concatenates strings.

src/components/TextInput/TextInput.tsx
oclero: The `getClassNames` function is available

src/components/TextInput/TextInput.tsx
pducolin: Changed by Quentin's PR (merged here now)

tsconfig.json
pducolin: Oh cool!

doczrc.js
pducolin: done by Quentin

tsconfig.json
pducolin: Can I put back `allowSyntheticDefaultImports`? Is there's anything against using synthetic default import?

src/components/TextInput/TextInput.tsx
pducolin: done by Quentin

src/typography/_typography.scss
pducolin: I Figma I only saw `Public Sans`, will check again.

src/utility/cssUtils.ts
pducolin: `classNames` package is this 30 lines of code https://github.com/JedWatson/classnames/blob/master/index.js#L13-L45

I don't mind adding it back if devs are more familiar, I agree with Olivier that such a small function can be an internal utility vs an external dependency

tsconfig.json
maryiam: I think the question is more: why do we need it ?

src/utility/cssUtils.ts
maryiam: I vote for no dependencies. We should keep this lib as light as possible :)

src/typography/_fonts.scss
```diff
@@ -0,0 +1,99 @@
+/* GT Walsheim Pro */
+@font-face {
```
maryiam, line 2: If a project doesn’t need all these fonts or wants to only target a font with a font-weight he must fetch all fonts files anyway? @adelego you had this issue with `@dashlane/fonts` [fonts file](https://gitlab.dashlane.com/webproduct/fonts/-/blob/master/src/index.css) no?

package.json
```diff
@@ -48,9 +50,12 @@
     "@types/styled-components": "^4.1.18",
     "@typescript-eslint/eslint-plugin": "^1.13.0",
     "@typescript-eslint/parser": "^1.13.0",
+    "autoprefixer": "^9.7.5",
     "awesome-typescript-loader": "^5.2.1",
     "cross-env": "^6.0.3",
+    "css-loader": "^3.4.2",
```
maryiam, line 56: I don't think we still need `css-loader`

src/utility/cssUtils.ts
oclero: If the team prefers to use the npm package, we can respect the team habits/preferencies, though.

Note that my small function does not cover as many types and combinations of types that the npm package covers. But do we really need to be able to handle nested arrays and objects?

maryiam: I think that the function covers the need for now. If we figure out later that we need to handle some other cases, we will adapt the code depending on the need!

oclero: That's how I see things :)

src/typography/_fonts.scss
pducolin: this file is meant to be used only inside this project. If we want to have it used by other projects it needs to be exported in `.css`. I have no idea about the fetching of fonts, need to spike on this.

maryiam: Since we import all fonts in `_typography.scss`, I think that we should check what's the final result in the bundle? In the docz build (in `docs`), I see that all fonts were included.

yarn.lock
```diff
@@ -1003,6 +1735,11 @@
   dependencies:
     find-up "^4.0.0"
 
+"@dashlane/fonts@^1.1.0":
+  version "1.1.0"
+  resolved "https://npm.dashlane.com/repository/npm/@dashlane/fonts/-/fonts-1.1.0.tgz#32f125f3ace7e73acfd284e55fa0c6a9ae8746b6"
```
maryiam, line 1740: We have Dashlane Nexus NPM private registry URL revealed here. Maybe we should let clients handle the install of fonts on their side for now? (All WP projects do it already)

pducolin: Counter part is that we don't have proper font rendering on docz

pducolin: Removing it!

pducolin: We should 
A: move fonts to github
B: move ui-components to gitlab

maryiam: I would say move ui-components to Gitlab. 
Also, I think that the whole fonts logic should be rethought again and we need to figure out how we can adapt fonts with ui-components without forcing clients to fetch all fonts.

tsconfig.json
pducolin: To be able to write `import React from 'react'` instead of `import * as React from 'react'` as far as I know

pducolin:
I have conflicts on artifacts (docs), any guideline?

yarn.lock
sericaia: is this blocking right now?

maryiam:
> I have conflicts on artifacts (docs), any guideline?

Remove all the folder and run docz build again.

yarn.lock
pducolin: I removed it, not blocking IMO to have docz not compliant because of fonts

pducolin: Out of curiosity, which is the part of the URL that should not be revealed?

.gitignore
```diff
@@ -5,3 +5,4 @@
 yarn-error.log
 .DS_Store
 docs
+*.css.map
```
maryiam, line 8: This one can be removed as well :) Sorry for the multiple comments

src/typography/_fonts.scss
pducolin: Is this a blocker?

yarn.lock
maryiam: The first part indicating the URL of our Nexus and then the name of registry like it's on Nexus, followed by the name of the package

src/typography/_fonts.scss
maryiam: Since we removed the font lib now, it's no longer a blocker. But, as I said in the other comment, in a more general way, we should take the time to rethink our fonts.

PR #29, opened 2020-03-25
Feature/update docz to 2.2

ni0ki:
This PR aims to update DocZ to [latest stable version 2.2](https://github.com/doczjs/docz/blob/master/CHANGELOG.md#220-2019-12-11)

Quite a lot changed under the hood, especially with the use of Gatsby to generate the resulting doc.
That means small configuration tweakings in doczrc.js and the new gatsby-node.js

Some problems still remain, such as the impossibility to style the <Playground>, which is fixed in 2.3 but the version is still in Alpha. Fix is for now to use an enclosing styled <div> as the root in the Playground.

Thanks for the reviews !


ni0ki:
For the record, this PR has been replaced by the one upgrading to DocZ 2.3 (https://github.com/Dashlane/ui-components/pull/54)

PR #30, opened 2020-03-25
Feature/alert

sericaia:
Adds the `Alert` component supporting `success` and `error` messages. Also adds two icons that were necessary for the alert: `CrossCircleIcon` and `CheckCircleIcon`.

I'll change the base repo to master as soon as `feature/textinput-css` is merged.

image


sericaia:
@maryiam @pducolin why do we need to version docz changes?  ([the update docz commit](https://github.com/Dashlane/ui-components/pull/30/commits/7f3bfdfa50b550002376944ee744754e91b22b88))

src/atoms/icons/CheckCircleIcon.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import SvgWrapper from '@ions/icon/wrapper/SvgWrapper';
+
+const CheckCircleIcon: React.FunctionComponent<
+  import('../../ions/icon/types').IconProps
+> = (props: import('../../ions/icon/types').IconProps): JSX.Element => (
```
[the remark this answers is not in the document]
pducolin: We slightly changed it - and need to move it here - but icons will become like [this](https://gitlab.dashlane.com/pducolin/ui-component-proposition/-/blob/master/src/components/Icon/SvgIcons/ArrowDownIcon.tsx)

`import React from "react";
import { SvgWrapper, IconProps } from "./components/SvgWrapper";

const ArrowDownIcon: React.FC<IconProps> = props => (
  <SvgWrapper {...props}>
    <React.Fragment>
      <path
        fillRule="evenodd"
        clipRule="evenodd"
        d="M10 11.5503L5.63659 6L4.5 7.00399L10 14L15.5 7.00399L14.3634 6L10 11.5503Z"
      />
    </React.Fragment>
  </SvgWrapper>
);

export default ArrowDownIcon;
`

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
...
+  children: string;
+}
+
+export const Alert: React.FC<AlertProps> = ({
+  severity = Severity.SUCCESS,
+  showIcon = true,
+  showCloseIcon = false,
+  children
+}) => {
+  const [close, setClose] = React.useState(false);
```
pducolin: Components should avoid having a state as it is considered an anti-pattern. To be discussed if it makes sense to have this one so that the alert is always closed on click, or if it should be the parent responsible for closing it, passing a `onClose` function through props

fmatias-dashlane: I'm used to working with design systems that provide both a Stateless and a Stateful version of these kinds of components to let developers decide what works best for their parent components. In the future maybe we should strive for this.
In this particular case I believe the Stateful component is the best default and I don't really see the need to implement a Stateless version right away. Thoughts?

pducolin: Oh I like the stateless/statefull option. Agree there should be a default one, I'm ok with this one!

sericaia: I've seen both too. In our use case I actually have another wrapper that also has to close it (after 5seconds). But that logic is a bit dependent on our use case and might not make sense to be here (at least fro now).

fmatias-dashlane: Yes, it's business logic dependent so it's a use case for a Stateless component.

src/atoms/icons/CheckCircleIcon.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
```
maryiam, line 1: Can you please avoid defining manually the icons?  Icons should be generated from design-token.
The process is described [here](https://github.com/Dashlane/ui-components/blob/master/CONTRIBUTING.md#good-to-know-) in the contributing.md in the last bullet point. You need to modify the design token (json) of icons and run a command that generates icons for you (reading from desing-token).

src/components/Alert/Alert.spec.tsx
```diff
@@ -0,0 +1,93 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+
+import { Alert, AlertProps, AlertSeverity } from './Alert';
+import { CheckCircleIcon, CloseIcon } from '@atoms';
+
+describe('<Alert />', () => {
+  let wrapper: any;
```
maryiam, line 8: Can we better type the `wrapper`?

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { Button, CheckCircleIcon, CloseIcon, CrossCircleIcon } from '@atoms';
+
+import './Alert.css';
```
maryiam, line 4: Why don't we simply work with sass files?

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const [close, setClose] = React.useState(false);
+
+  if (!children || close) {
+    return null;
+  }
+
+  const renderIcon = (severity: AlertSeverity) => {
+    let icon;
+
+    switch (severity) {
```
maryiam, line 33: I'm not really fun of mutating variables as it's anti-pattern. Can you maybe extract the logic from `switch` and move it into a function `getIconBySeverity` for example?

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
...
+        icon = <CheckCircleIcon size={30} color='white' />;
+        break;
+      }
+    }
+
+    return <div className='icon'>{icon}</div>;
+  };
+
+  return (
+    <div className={'root ' + severity}>
```
maryiam, line 49: I think that we added lately some `classname` helper, can we use it here?

src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { Button, CheckCircleIcon, CloseIcon, CrossCircleIcon } from '@atoms';
+
+import './Alert.css';
+
+export enum AlertSeverity {
```
maryiam, line 6: Can we stick to the current convention of declaring types that will be useful for clients in a separate file called `types.ts` like [in this example](https://github.com/Dashlane/ui-components/blob/master/src/atoms/button/types.ts)

maryiam:
> @maryiam @pducolin why do we need to version docz changes? ([the update docz commit](https://github.com/Dashlane/ui-components/pull/30/commits/7f3bfdfa50b550002376944ee744754e91b22b88))

Well, since we don't have a CI running on this project, we need to update the doc ourselves so it can be auto-deployed on Github pages [here](https://dashlane.github.io/ui-components/)

src/components/Alert/Alert.tsx
sericaia: I'm not mutating directly, I create a new variable `icon`, which is different depending on the input. However, I can extract the function, no prob

src/components/Alert/Alert.tsx
pducolin: could the timer based closing be a prop? (to be discussed with design). We had a similar component on UWP

pducolin: Don't know if you can access. It's C#. We pass an `ICommand` (equivalent to a function) that is called back on click. But I don't know if it works well here as we rerender the component on prop change.

https://gitlab.dashlane.com/windows/windows-uwp-client/-/blob/master/windows/Frameworks/DesignSystem/DesignSystem.UWP/Molecules/DLInAppNotification.xaml.cs

fmatias-dashlane: That's a good idea, with a `closeDelay` prop (default null) we can take the closing delay logic outside of the client.
Nevertheless, that feels like an improvement on Daniela's code and that could be done in another PR to unblock stories that depend on this PR. Do you agree with that approach?

sericaia: I agree with Fabio as this is delaying us and that's an improvement.

pducolin: oh yes sure! Reviewing the PR

PR #31, opened 2020-03-31
Feature/move to components

pducolin:
Moving tooltip and dropdown to components


tsconfig.json
```diff
@@ -20,6 +20,8 @@
       "@ions/*": ["ions/*"],
       "@icons": ["icons/index"],
       "@icons/*": ["icons/*"],
+      "@components": ["components/index"],
```
adelego: What is the difference between a component and an atom or molecule?

pducolin: Components has one less level of abstraction. The atom architecture works very well for the conception of a design system (helps asking the question if one element should be an ions vs an atom, and de structuring molecules in sub-elements), but for the implementation can cause friction, because the library user has to have prior knowledge of those concepts - and if they are not familiar, they can get lost.

So for DS implementation it is common to use a single folder - components - where all the elements can be found. There is still difference between components that are both exported and used in other components (for example button), and those which are only used in a single components (for example DropDownCard). These should be isolated in a `subcomponents` folder and be accessible only by the parent element.

We used the atomic architecture for Windows, but new arrivals had friction because of the abstraction. During the workshop at DS London with Jina Anne she also suggested to avoid the atomic architecture in the implementation. Can't find any article with more details but will keep looking.

oclero: TL;DR The classification ion/atom/molecule/organism is an eternal debate, and as developers or designers we don't have the same vision about what's a ion, what's an atom, what's a molecule. Because this classification has no concrete technical universal meaning, it's much simpler to put everything into "components".

pducolin:
Strangely enough my commits are not signed, will rebase and sign them

tsconfig.json
maryiam: Personally, I think that even if we stop splitting with atoms/ions logic,  it still makes sense to split components into two main categories: "components to be used by clients" and "internal components to ui-components lib".

maryiam:
Why don't we move the rest of the components?

pducolin:
> Why don't we move the rest of the components?

I will, I preferred doing it in 2 PRs to keep as few changes as possible in each one.

tsconfig.json
pducolin: We are still splitting the lib this way - internal components are isolated in `subcomponents` folder.

maryiam: Oh ok great! I thought that everything is going to be moved to `components`! :)

PR #33, opened 2020-03-31
feat: scss modules support complete

ni0ki:
Added clean SCSS modules / support to docz, jest, and the build process.
Updated the TextInput component to reflect the changes.

The styles are now properly scoped to their component, with no collision risk. The resulting CSS is embedded in the final JS file, and is injected at runtime.


sericaia:
This looks good, thanks. The only thing I would change is to do 2 different PRs next time. One related wits SCSS and another for the TextInput ;)

jest.config.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
   roots: ['<rootDir>/src'],
   transform: {
+    '.+\\.(css|styl|less|sass|scss)$':
+      '<rootDir>/node_modules/jest-css-modules-transform',
```
pducolin: Isn't enough to give it `jest-css-modules-transform` instead of the full path with `<rootDir>` ?

package.json
```diff
@@ -26,19 +26,19 @@
     "check-lint": "eslint --ext tsx,ts,js,json .",
     "exec-docz-update": "node config/updateDoczBuild.js",
     "generate-icon-components": "node config/generator/icons/generateIcons.js",
-    "generate-colors-stylesheets": "node config/generator/colors/generateColorStylesheets.js",
-    "sass:watch": "yarn sass --watch src:src"
+    "generate-colors-stylesheets": "node config/generator/colors/generateColorStylesheets.js"
   },
   "sideEffects": false,
   "dependencies": {
+    "@babel/runtime": "^7.9.2",
```
pducolin: this one can be removed

src/components/TextInput/TextInput.tsx
```diff
@@ -18,38 +17,28 @@ const classNameFromFeedbackType: Record<TextInputFeedbackType, string> = {
   warning: 'warning'
 };
 
-const getClassNameFromFeedbackType = (
-  feedbackType?: TextInputFeedbackType
-): string => {
+const getClassNameFromFeedbackType = (feedbackType?: TextInputFeedbackType) => {
```
pducolin: as Mariem pointed out in my PR, we can refactor this and use directly the prop instead of converting it to text - I will change it later

src/components/TextInput/TextInput.tsx
```diff
@@ -13,43 +12,27 @@ interface TextInputProps
...
-    feedbackText.length > 0
-  );
+const shouldShowFeedback = ({ feedbackText, feedbackType }: TextInputProps) => {
+  return feedbackText && feedbackType;
 };
 
 export const TextInput: React.FC<TextInputProps> = props => {
-  const className = getClassNameFromFeedbackType(props.feedbackType);
-  const { feedbackText, feedbackType, ...htmlInputProps } = props;
+  const styledClassName = props.feedbackType
```
maryiam, line 20: I don't find this nested ternary clear and easily understandable.
maybe 
```js
props.feedbackType && style[props.feedbackType] ? style[props.feedbackType] : ''
```

src/components/TextInput/TextInput.module.scss
```diff
@@ -63,19 +63,19 @@
     &::-ms-input-placeholder {
```
maryiam, line 63: We should have a linter for styles to avoid this kind of changes in PRs

jest.config.js
maryiam: yes, I think you can omit the `<rootDir>/node_modules/` part

jest.config.js
```diff
@@ -14,7 +16,6 @@ module.exports = {
     '^@ions/(.*)$': '<rootDir>/src/ions/$1',
     '@ions': '<rootDir>/src/ions/index',
     '@colors': '<rootDir>/src/colors/colors.ts',
-    '@colors/*': '<rootDir>/src/colors/*',
-    '\\.(css|less)$': 'identity-obj-proxy'
```
maryiam, line 18: If `identity-obj-proxy` is not used any more than maybe we can delete it from our dependencies?

package.json
```diff
@@ -82,13 +85,14 @@
     "rollup-plugin-terser": "^5.1.2",
     "rollup-plugin-typescript2": "^0.24.3",
     "sass": "^1.26.3",
+    "sass-loader": "^8.0.2",
     "semantic-release": "^15.13.27",
     "style-loader": "^1.1.3",
```
maryiam, line 90: Is `style-loader` still needed?

package.json
```diff
@@ -38,7 +37,7 @@
   },
   "devDependencies": {
     "@ampproject/rollup-plugin-closure-compiler": "^0.10.0",
-    "@babel/runtime": "^7.9.2",
+    "@callstack/react-theme-provider": "^2.1.0",
```
maryiam, line 40: Why is this needed?

src/typings.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module '*.module.scss' {
```
maryiam, line 1: Isn't it enough to declare the module `*.module.scss`? Why is `className` needed?

src/typings.d.ts
```diff
@@ -0,0 +1,7 @@
...
+  interface ClassNames {
+    [className: string]: string;
+  }
+  const classNames: ClassNames;
```
maryiam, line 5: You don't need an interface here, you can use [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt)
```js
classNames: Record<string, string>
```

src/typings.d.ts
```diff
@@ -0,0 +1,7 @@
...
+  export = classNames;
```
maryiam, line 6: Or export default classNames?

jest.config.js
ni0ki: Will do

src/components/TextInput/TextInput.tsx
ni0ki: Will do

jest.config.js
ni0ki: Sure, will do

package.json
ni0ki: Nope, I'll remove it

package.json
ni0ki: Because we mixed up things in this TextInput PR, namely, it was the original one to provide SASS support, which was needed for the theming we started, which relies on this library. So everything related to SASS / PostCSS / theming is based on this branch.

src/typings.d.ts
ni0ki: Yes, it is enough. I was originally using another rollup-postcss plugin which was causing some TS problems, with the changes the className part is indeed not needed :+1:

src/typings.d.ts
ni0ki: Will be deleted !

src/components/TextInput/TextInput.tsx
ni0ki: I changed it !

package.json
maryiam: Maybe remove temporarily then? :)

src/components/TextInput/TextInput.module.scss
oclero: I'm adding the necessary lines to .prettierrc in another PR.

maryiam: That's cool :) Thanks

PR #35, opened 2020-04-01
Fix: fix manually generated family icons

fmatias-dashlane:
The family icons (logo, settings and link) were generated manually and weren't included in `src/design-tokens/icons/icons-defs.json`. This PR fixes that.

The Dashlane logo was also moved to `src/components/` to separate it from the other regular 20x20 square icons that are automatically generated.


src/atoms/icons/LogoIcon.tsx
```diff
@@ -1,14 +1,17 @@
...
-    />
+const LogoIcon: React.FunctionComponent<
+  import('../../ions/icon/types').IconProps
+> = (props: import('../../ions/icon/types').IconProps): JSX.Element => (
+  <SvgWrapper {...props}>
```
fmatias-dashlane, line 7: @maryiam This icon isn't square and needs to have an explicit `viewBox` value.
Do you think I should edit this manually after the icon is automatically generated? Or should there be another change of any kind?

maryiam: @fmatias-dashlane AFAIK all web Icons are supposed to be set with a viewBox of `0 0 20 20`. Isn't this case for your icon?

fmatias-dashlane: This is the Dashlane logo icon, it's a rectangle.
image

fmatias-dashlane: Should it not be an icon?

maryiam: Ah I see, I think that it should be an icon. The problem here if you are going to modify `src/atoms/icons/LogoIcon.tsx` it will be overridden the next time someone else runs `yarn generate-icon-components`. If it's a particular case (we only have this icon with a different viewbox) maybe let the client pass the `viewBox` prop ? Otherwise, if we consider that we will have many icons in the future with bigger or smaller viewBox, then maybe we should adapt the current icon-generation behaviour to take into account different viewBoxes.

pducolin: You can also check with your designer. We can consider the `icon` component to only be squared. If this is the case, the business logo could be kept out of the DS.

fmatias-dashlane: I disagree with letting the client handle the viewBox prop.
This being a design system, the client shouldn't be concerned with the Logo viewBox, all a client cares about is having a logo rendered with the proper aspect ratio.

Is there really a lot to gain from having the icons automatically generated?
It was pretty simple to check how the other icons were implemented and replicate that. Having this automatically done raised this viewBox issue and will probably raise other issues in the future.
The only alternative I see is to not consider the logo an icon 🤷‍♂

maryiam: The gain with icons generation is seen when editing old icons. Lately, designers have been changing quite frequently the icons. While keeping the icons in a design token, we will be able to create a Figma plugin to change the icons in ui-components for each design change with a single command line. We did this already with the colors: 

- Designer changes colors in Figma
- Designer uses the Figma plugin to apply color changes in ui-components
- The plugin will handle updating the color design token json file + creates a Pull request with the change
- We simply re-run the command line `yarn generate-colors-stylesheets` to apply all new changes.

fmatias-dashlane: I also don't agree with having the logo outside the DS 😛 
We need to keep a consistent UI across our products and we're currently experiencing many inconsistency pains and one of them is having several different versions of the logo in multiple projects (and also multiple on the same project).
With the logo living inside the DS and every product using that one, everything will be consistent. Think of the rebranding effort that would save, for example 😄

pducolin: Agree on that. Then I would have it as another component - BrandIcon for example, that can be any size we want. This would keep icons squared (in all meanings)

pducolin: To @maryiam point, I would keep the manual extraction as long as we don't have full process automation.

I see the gain of extracting all icons in one click from Figma, what I find misleading is having both sourcecode and generated artifacts tracked by the repo. I would rather build them as a pre-build step - same for colors and typography. What do you think?

pducolin: @fmatias-dashlane this kind of decision is best taken with designers - this is the power of DS, we all set and follow the same rules

maryiam: @pducolin I don't fully understand your point. What do you mean by pre-build step? Don't forget that there is a lot of constraints currently due to lack of CI/CD on the project, once we will have that we will adapt/facilitate most of the verification/hooks that we have now. Honestly, I don't vote for manual process/actions, because once you start doing that, it's hard to go back later on.
Otherwise, a logo component is fine with me!

fmatias-dashlane: Ok, so I'll have the logo separated from the icons :)
Any suggestions where I should move it to? Maybe `src/atoms/logo/`?

pducolin: You can put it in `src/components` (rebase on master and you'll have the new folder)

pducolin: @maryiam what I mean is that the current process has more steps than the manual one, so devs have more friction. 

I see the value in having an icon generator, but probably more as a one-time generator (thinking of [svg to jsx tool](https://svg2jsx.com/)) - it could take as an input the svg, and create the jsx by removing the `svg` tags and escaping th `"`.

What is the use case that would cause issues once we fully automate the process? From my point of view we can keep tracking in git the icons,  and once we have CI/CD in place we can stop tracking them removing them and ignoring them (anything mean we can do to them through git 👿 ) and generating them directly from the json data source, that is generated by a Figma plugin. Any opinion?

fmatias-dashlane: 200% agree with the one-time tool 👏 That way the icon would get generated and could be customized without worrying that another automatic generation would override those changes. Nevertheless, it's a question of deciding what we prefer. If we stick to all icons being a square icon jsx coming from Figma exported SVGs, the current approach works 🤷‍♂ 

Regarding the Figma-CI-Git discussion, I believe Git should be the single source of truth to allow deploying only what was tested. If we choose to adopt Figma as the source of truth, changes to it should be versioned and "released" (I have no idea if Figma supports this).

Anyhow, I think we can all agree this discussion has gone out of this PR's scope 😛 
Let's schedule a meeting to discuss this further if necessary 🙏

PR #36, opened 2020-04-01
feat: use unified tabWidth for prettier autoformat

oclero:
We need a unified auto-formatting, especially for tabs/spaces in CSS. Otherwise we'll get undesired line changes in commits. Let's do it before the repository grows too much.

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,6 +1,4 @@
 module.exports = {
   singleQuote: true,
-  jsxSingleQuote: true,
-  tabWidth: 2,
-  useTabs: false
+  jsxSingleQuote: true
 };
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,4 +1,6 @@
-module.exports =  {
-    singleQuote:  true,
-    jsxSingleQuote: true
+module.exports = {
+  singleQuote: true,
+  jsxSingleQuote: true,
+  tabWidth: 2,
+  useTabs: false
 };
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,11 +103,17 @@ All you need to do is :
 
 ### Code linting and formatting
 
-We use `prettier` to format and `eslint` with typescript settings to lint our code.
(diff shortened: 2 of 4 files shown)


.prettierrc.js
```diff
@@ -1,4 +1,6 @@
-module.exports =  {
-    singleQuote:  true,
-    jsxSingleQuote: true
+module.exports = {
+  singleQuote: true,
+  jsxSingleQuote: true,
+  tabWidth: 2,
```
maryiam, line 4: I think that `tabWidth` and `useTabs` have these values (`2` and `false`) by default?
https://prettier.io/docs/en/options.html

oclero: Ha yes you're right. Then... how can people commit files with 4 spaces?

maryiam: Maybe it's because of the [`editorConfig`](https://editorconfig.org/) ?

oclero: We could use that indeed. I can add it in this PR.

CONTRIBUTING.md
```diff
@@ -103,11 +103,17 @@ All you need to do is :
...
 ### Good to know :
 
-- It's convenient to configure your ide in order automatically format your files. Please follow this [Editor Integration documentation](https://prettier.io/docs/en/editors.html).
+- It's convenient to configure your IDE in order to automatically format your files. Please follow this [Prettier Editor Integration documentation](https://prettier.io/docs/en/editors.html).
+
+- To ensure your IDE is already correctly configured, we use [EditorConfig](https://editorconfig.org/). Please install the required plugin for it, if your IDE needs one:
+
+  - [VSCode plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig)
+  - [Vim plugin](https://github.com/editorconfig/editorconfig-vim)
+  - WebStorm supports it natively.
```
maryiam, line 116: Apparently there is one https://plugins.jetbrains.com/plugin/7294-editorconfig 😛 (But it comes installed by default with webstorm)

PR #41, opened 2020-04-07
Feature/password input

a-coruble:
One thing to note, is that some design decisions were not implemented (per-character text color, last input visible while password hidden and typing)


a-coruble:
Error with rebase, will publish a new branch and open a new PR

PR #43, opened 2020-04-07
[WIP] feature/import optimization

ni0ki:
This PR changes how the library is exported. Before, we'd put everything in a single index.esm.js, now we preserve the modular structure used in src/

This should not change the way to consume it, as it's still exported within the ES module format.

Big advantage is that we won't bundle the style of unused components like we did before. This was because CSS modules inject CSS-in-JS @ runtime. Now, this injection stays "scoped" in its component, so it won't be injected if the component is not imported.

I had to change the minifier used as the closure compiler was not able to follow that modular structure. Terser is widely used and seems on par with it.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,34 +1,52 @@
 import postcss from 'rollup-plugin-postcss';
-import compiler from '@ampproject/rollup-plugin-closure-compiler';
+import { terser } from 'rollup-plugin-terser';
 import typescript from 'rollup-plugin-typescript2';
 import json from 'rollup-plugin-json';
 import pkg from './package.json';
-
+import renameExtensions from 'rollup-plugin-rename';
 import autoprefixer from 'autoprefixer';
+import replace from '@rollup/plugin-replace';
 
 const config = {
   input: 'src/index.ts',
-
+  preserveModules: true,
   external: [
     ...Object.keys(pkg.dependencies || {}),
     ...Object.keys(pkg.peerDependencies || {})
   ],
   plugins: [
+    /** SCSS transformation step. Compiles into CSS modules and minify outpyt */
     postcss({
       extract: false,
-      plugins: [autoprefixer()]
+      plugins: [autoprefixer()],
+      minimize: true
     }),
     typescript({
       typescript: require('typescript'),
       objectHashIgnoreUnknownHack: true
     }),
-    json({ preferConst: true })
+    json({ preferConst: true }),
+    /**
+     * Next lines are tricky.
+     * For unknown reasons, when the postcss extract option is set to false using preserveModules set to true,
+     * the .js files will be created in /lib/src
+     * This is not the wanted output, so we move them back into lib/ ,
+     * but we have one import that must be updated
+     */
+    replace({
+      '../_virtual/_tslib.js': '_virtual/_tslib.js',
+      delimiters: ['', '']
+    }),
+    renameExtensions({
+      include: ['**/*'],
+      map: name => name.replace('src/', '')
+    })
   ],
-  output: [{ file: pkg.module, format: 'esm', sourcemap: true }]
+  output: [{ dir: 'lib', format: 'esm', sourcemap: true }]
 };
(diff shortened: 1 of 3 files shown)


ni0ki:
I'd like someone other than me to test it inside the popup to make sure I didn't screw something up. Thanks !

rollup.config.js
```diff
@@ -1,34 +1,51 @@
...
 const config = {
   input: 'src/index.ts',
-
+  preserveModules: true,
   external: [
     ...Object.keys(pkg.dependencies || {}),
     ...Object.keys(pkg.peerDependencies || {})
   ],
   plugins: [
+    /** SCSS transformation step. Compiles into CSS modules and minify outpyt */
```
pducolin, line 18: s/outpyt/output

rollup.config.js
```diff
@@ -1,34 +1,51 @@
...
+      minimize: true
     }),
     typescript({
       typescript: require('typescript'),
       objectHashIgnoreUnknownHack: true
     }),
-    json({ preferConst: true })
+    json({ preferConst: true }),
+    /**
+     * When preserveModules is set to true, rollup mimics the file structure.
```
pducolin, line 30: Would add this line also before `preserveModules: true`

rollup.config.js
```diff
@@ -1,34 +1,51 @@
...
+     * Therefore, the .js files will be created in /lib/src. Move them back to /lib
```
pducolin, line 31: Would rephrase with `replace moves them back...`

rollup.config.js
```diff
@@ -1,34 +1,51 @@
 import postcss from 'rollup-plugin-postcss';
-import compiler from '@ampproject/rollup-plugin-closure-compiler';
```
maryiam, line 2: I'm sad that we are moving away from the closure compiler as it is a very powerful (it produces superior minification than other libs). But I understand that it might be hard to configure with multiple entries instead of a single one.

rollup.config.js
```diff
@@ -1,34 +1,51 @@
...
+     * This also makes some modules "virtual" and put them into a special folder.
+     * We therefore need to update the referenced path as well
+     */
+    replace({
+      '../_virtual/': '_virtual/',
+      delimiters: ['', '']
+    }),
+    renameExtensions({
+      include: ['**/*'],
+      map: name => name.replace('src/', '')
```
maryiam, line 41: This seems to be a very hacky solution. 
I tried to build locally and I see that we have now `src` level because rollup has two folder entries `node_modules` (for `inject-style` dep) and `src` (for the components). The downside of this is that the `inject-style` dep is injected in each component using Sass now. We can discuss this further if you would like to.

rollup.config.js
ni0ki: I was unable to have the closure compiler working with preserveModules unfortunately.
But on master, when I use terser instead of the closure compiler, I get a smaller bundler, so I guess this shouldn't be a problem ?

rollup.config.js
ni0ki: Good catch for the nodes_modules, I didn't see it !
I'll try to update this, but I don't know where to start

rollup.config.js
maryiam: Nice :)

goldins:
@ni0ki can this PR be updated or closed?

ni0ki:
@goldins 
TL;DR This PR is about a potential optimization. Not a huge end-user value, but it would need some time from senior people 

This would, I think, need to be discussed. I'm no expert in Webpack or tree-shaking, but because CSS modules export the styles as JS objects, I think the whole styles from every component is bundled into final app because they can't be properly tree-shaken (from what I can remember in what we tested, this would require more checking by someone with more experience than me).

Goal of this PR was to try a different bundling output (keeping the esm modules structure, not putting everything in a single file), so that style could be tree-shaken efficiently as well. It was working, solving that problem from what I could see, but could potentially cause other problems.

PR #44, opened 2020-04-07
Feature/correct password input

a-coruble:
Important to note that some design features have not been implemented (character specific color, last typed character visible with other hiddens,)

Here are the screenshots :
Screenshot 2020-04-09 at 11 12 26
Screenshot 2020-04-09 at 10 24 41

### DIFF
--- a/src/components/password-input/PasswordInput.mdx
+++ b/src/components/password-input/PasswordInput.mdx
@@ -5,7 +5,6 @@ menu: Components
 
 import { Playground, Props } from 'docz';
 import { PasswordInput } from './PasswordInput';
-import { PasswordStrength } from './subcomponents/PasswordStrength/PasswordStrength.tsx';
 
 # Password Input
 
@@ -15,10 +14,6 @@ A Password input component
 
 <Props of={PasswordInput} />
 
-#### Password Strength Props
-
-<Props of={PasswordStrength} />
-
 ## Usage
 
 ### Basic usage
@@ -27,61 +22,40 @@ Examples below show usage of the Button
 
 <Playground
   style={{
+    alignItems: 'center',
     display: 'flex',
-    flexDirection: 'row',
+    flexDirection: 'column',
     justifyContent: 'space-between',
     height: 400,
     flex: 1
   }}
 >
-  <div
-    style={{
-      display: 'flex',
-      flexDirection: 'column',
-      justifyContent: 'space-evenly'
-    }}
-  >
-    <PasswordInput placeholder='Test' />
-    <PasswordInput showPassword />
-    <PasswordInput disabled />
-    <PasswordInput feedbackType='warning' feedbackText='Test' />
-    <PasswordInput feedbackType='error' feedbackText='Test' />
-  </div>
-  <div
-    style={{
-      display: 'flex',
-      flexDirection: 'column',
-      justifyContent: 'space-evenly'
-    }}
-  >
-    <PasswordInput
-      placeholder='Test'
-      showPasswordStrength={true}
-      passwordStrengthProps={{
-        score: 0,
-        showAdditionalInfo: true,
-        additionalInfo: 'Some Test'
-      }}
-    />
-    <PasswordInput
-      placeholder='Test'
-      showPasswordStrength={true}
-      passwordStrengthProps={{ score: 1 }}
-    />
-    <PasswordInput
-      placeholder='Test'
-      showPasswordStrength={true}
-      passwordStrengthProps={{ score: 2 }}
-    />
-    <PasswordInput
-      placeholder='Test'
-      showPasswordStrength={true}
-      passwordStrengthProps={{ score: 3 }}
-    />
-    <PasswordInput
-      placeholder='Test'
-      showPasswordStrength={true}
-      passwordStrengthProps={{ score: 4 }}
-    />
-  </div>
+  <PasswordInput
+    hidePasswordTooltipText='Hide'
+    showPasswordTooltipText='Reveal'
+    label='Password'
+    placeholder='Test'
+  />
+  <PasswordInput
+    hidePasswordTooltipText='Hide'
+    showPasswordTooltipText='Reveal'
+    showPassword
+  />
+  <PasswordInput
+    hidePasswordTooltipText='Hide'
+    showPasswordTooltipText='Reveal'
+    disabled
+  />
+  <PasswordInput
+    hidePasswordTooltipText='Hide'
+    showPasswordTooltipText='Reveal'
+    feedbackType='warning'
+    feedbackText='This password contains invalid characters'
+  />
+  <PasswordInput
+    hidePasswordTooltipText='Hide'
+    showPasswordTooltipText='Reveal'
+    feedbackType='error'
+    feedbackText='Wrong password'
+  />
 </Playground>
--- a/src/components/password-input/PasswordInput.mdx
+++ b/src/components/password-input/PasswordInput.mdx
@@ -1,0 +1,87 @@
+---
+name: Password Input
+menu: Components
+---
+
+import { Playground, Props } from 'docz';
+import { PasswordInput } from './PasswordInput';
+import { PasswordStrength } from './subcomponents/PasswordStrength/PasswordStrength.tsx';
+
+# Password Input
+
+A Password input component
+
+## Properties
+
+<Props of={PasswordInput} />
+
+#### Password Strength Props
+
+<Props of={PasswordStrength} />
+
+## Usage
+
+### Basic usage
+
+Examples below show usage of the Button
+
+<Playground
+  style={{
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-between',
+    height: 400,
+    flex: 1
+  }}
+>
+  <div
+    style={{
+      display: 'flex',
+      flexDirection: 'column',
+      justifyContent: 'space-evenly'
+    }}
+  >
+    <PasswordInput placeholder='Test' />
+    <PasswordInput showPassword />
+    <PasswordInput disabled />
+    <PasswordInput feedbackType='warning' feedbackText='Test' />
+    <PasswordInput feedbackType='error' feedbackText='Test' />
+  </div>
+  <div
+    style={{
+      display: 'flex',
+      flexDirection: 'column',
+      justifyContent: 'space-evenly'
+    }}
+  >
+    <PasswordInput
+      placeholder='Test'
+      showPasswordStrength={true}
+      passwordStrengthProps={{
+        score: 0,
+        showAdditionalInfo: true,
+        additionalInfo: 'Some Test'
+      }}
+    />
+    <PasswordInput
+      placeholder='Test'
+      showPasswordStrength={true}
+      passwordStrengthProps={{ score: 1 }}
+    />
+    <PasswordInput
+      placeholder='Test'
+      showPasswordStrength={true}
+      passwordStrengthProps={{ score: 2 }}
+    />
+    <PasswordInput
+      placeholder='Test'
+      showPasswordStrength={true}
+      passwordStrengthProps={{ score: 3 }}
+    />
+    <PasswordInput
+      placeholder='Test'
+      showPasswordStrength={true}
+      passwordStrengthProps={{ score: 4 }}
+    />
+  </div>
+</Playground>
--- a/src/components/password-input/PasswordInput.mdx
+++ b/src/components/password-input/PasswordInput.mdx
@@ -35,11 +35,11 @@ Examples below show usage of the Button
     showPasswordTooltipText='Reveal'
     label='Password'
     placeholder='Test'
+    onPasswordVisibilityChanged={nextState => console.log(nextState)}
   />
   <PasswordInput
     hidePasswordTooltipText='Hide'
     showPasswordTooltipText='Reveal'
-    showPassword
   />
   <PasswordInput
     hidePasswordTooltipText='Hide'
--- a/src/components/password-input/PasswordInput.module.scss
+++ b/src/components/password-input/PasswordInput.module.scss
@@ -60,7 +60,7 @@
       padding-right: 12%;
 
       &:focus {
-        border: 1px solid colors.$mid-green-01;
+        border-color: colors.$mid-green-01;
       }
 
       &:disabled {
@@ -69,21 +69,25 @@
       }
 
       &.warning {
-        border: 1px solid colors.$orange-00;
+        border-color: colors.$orange-00;
       }
 
       &.error {
-        border: 1px solid colors.$functional-red-02;
+        border-color: colors.$functional-red-02;
       }
     }
 
     .icon {
       position: absolute;
       top: 10px;
-      right: 14px;
+      right: 16px;
       background: none;
       border: none;
       padding: 0;
+
+      &:focus {
+        outline: none;
+      }
     }
   }
 }
--- a/src/components/password-input/PasswordInput.module.scss
+++ b/src/components/password-input/PasswordInput.module.scss
@@ -81,6 +81,9 @@
       position: absolute;
       top: 10px;
       right: 14px;
+      background: none;
+      border: none;
+      padding: 0;
     }
   }
 }
--- a/src/components/password-input/PasswordInput.module.scss
+++ b/src/components/password-input/PasswordInput.module.scss
@@ -1,0 +1,86 @@
+@use '../../colors/colors';
+@use '../../typography/typography';
+
+/* remove edge decorators */
+::-ms-clear {
+  display: none;
+}
+
+.root {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+
+  .label {
+    @include typography.body-small;
+
+    color: colors.$dash-green-01;
+    text-transform: uppercase;
+    height: 20px;
+    font-size: 10px;
+    line-height: 20px;
+    font-weight: 600;
+    margin-bottom: 4px;
+  }
+
+  .feedback {
+    @include typography.body-small;
+
+    &.warning {
+      color: colors.$orange-00;
+    }
+
+    &.error {
+      color: colors.$functional-red-02;
+    }
+  }
+
+  .container {
+    position: relative;
+    margin-bottom: 4px;
+
+    .input {
+      outline: none;
+
+      /* typography */
+      @include typography.body-medium;
+
+      /* colors */
+      color: colors.$black;
+      background: colors.$white;
+      height: 40px;
+      width: 300px;
+
+      /* border */
+      border: 1px solid colors.$dash-green-05;
+      box-sizing: border-box;
+      border-radius: 4px;
+
+      padding-left: 16px;
+      padding-right: 12%;
+
+      &:focus {
+        border: 1px solid colors.$mid-green-01;
+      }
+
+      &:disabled {
+        background: colors.$dash-green-06;
+        cursor: not-allowed;
+      }
+
+      &.warning {
+        border: 1px solid colors.$orange-00;
+      }
+
+      &.error {
+        border: 1px solid colors.$functional-red-02;
+      }
+    }
+
+    .icon {
+      position: absolute;
+      top: 10px;
+      right: 14px;
+    }
+  }
+}
--- a/src/components/password-input/PasswordInput.tsx
+++ b/src/components/password-input/PasswordInput.tsx
@@ -1,0 +1,153 @@
+import * as React from 'react';
+
+import { getClassNames } from '@utility/cssUtils';
+import { HideIcon, RevealIcon } from '@atoms/icons';
+import { midGreen00 } from '@colors';
+import { Tooltip } from 'components';
+
+import {
+  PasswordStrength,
+  PasswordStrengthProps
+} from './subcomponents/PasswordStrength/PasswordStrength';
+import styles from './PasswordInput.module.scss';
+
+type PasswordInputFeedbackType = 'error' | 'warning';
+
+type _ForbiddenProps = 'size' | 'prefix' | 'type';
+
+export interface PasswordInputProps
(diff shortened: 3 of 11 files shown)


adelego:
Hello @a-coruble ! Can you provide screenshots of the input please?

src/components/password-input/PasswordInput.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+# Password Input
+
+A Password input component
+
+## Properties
+
+<Props of={PasswordInput} />
+
+#### Password Strength Props
```
adelego, line 18: Can you elaborate on what is the component PasswordStrength for?

adelego: Also, there should be a dedicated doc page for this as someone may want to use PasswordInput without the strength

src/components/password-input/PasswordInput.module.scss
```diff
@@ -0,0 +1,86 @@
+@use '../../colors/colors';
```
adelego, line 1: We will need a specific style for dark mode (no need to include it in this PR of course). @pducolin @maryiam how do you us doing that with sass? Is there a specific way or will we use class names?

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+  label?: string;
+  /**
+   * A function that controls the behavior of the show / hide feature.
+   * You should use that with the `showPassword` property to make a controlled component
+   */
+  onShowHidePassword?: () => void;
+  /**
+   * Props for the password strength indicator, see below for better description.
+   */
+  passwordStrengthProps?: PasswordStrengthProps;
```
adelego, line 48: Because the libs user will need PasswordInput without passwordStrength and vice versa, I would suggest making two standalone components. This only setback we can't make sure that the developers uses the same "strength" value

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * Boolean triggering the display / hide of the password
+   * @default false
+   */
+  showPassword?: boolean;
```
adelego, line 53: If `showPassword` is defined you must have an `onShowPassword` props as well. We can improve typing by putting both in an interface

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+    showPasswordTooltipText = 'Reveal',
+    ...rest
+  } = props;
+
+  const [showPasswordState, setShowPasswordState] = React.useState(
+    showPassword
+  );
+
+  React.useEffect(() => {
+    setShowPasswordState(showPassword);
```
adelego, line 100: I don't like that we duplicate the prop in the component's state when `props.showPassword` is defined. 
Can we simplify this by:
- letting this component handle showPassword and setShowPassword and removing showPassword from the props,
- having a props onShowPasswordCallback that is called in the click handler where users can for instance send logs ?

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+  };
+
+  const styledClassName =
+    feedbackType && styles[feedbackType] ? styles[feedbackType] : '';
+
+  const Icon = showPasswordState ? HideIcon : RevealIcon;
+
+  return (
+    <div className={styles.root}>
+      <span className={styles.label}>{label}</span>
```
adelego, line 118: What happens when `label` is undefined?

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+}: PasswordInputProps) => {
+  return showPasswordStrength && passwordStrengthProps;
+};
+
+export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
+  const {
+    feedbackText,
+    feedbackType,
+    hidePasswordTooltipText = 'Hide',
+    label = 'Password',
```
adelego, line 86: label should not be defined by default. I can think of designs where we don't use it

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+  passwordStrengthProps
+}: PasswordInputProps) => {
+  return showPasswordStrength && passwordStrengthProps;
+};
+
+export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
+  const {
+    feedbackText,
+    feedbackType,
+    hidePasswordTooltipText = 'Hide',
```
adelego, line 85: We don't want to let developers not define this as it should be properly translated

adelego: Same for showPasswordTooltipText

src/components/password-input/subcomponents/PasswordStrength/PasswordStrength.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import { getClassNames } from '@utility/cssUtils';
+
+import styles from './PasswordStrength.module.scss';
+
+export interface PasswordStrengthProps {
+  /**
+   * Any text you want to display under the strength indicator
+   */
+  additionalInfo?: string;
```
adelego, line 11: Info is not specific to me. Why not additionalText?

src/components/password-input/PasswordInput.module.scss
pducolin: We were just talking about theming during the standup, @ni0ki and @oclero did an investigation on theming and they will prepare an example or guide to use it. We will use the css `prefers-color-scheme` to switch between dark and light. More details to come.

src/components/password-input/PasswordInput.tsx
pducolin: Indeed this looks like https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    <div className={styles.root}>
+      <span className={styles.label}>{label}</span>
+      <div className={styles.container}>
+        <input
+          {...rest}
+          type={showPasswordState ? 'text' : 'password'}
+          className={getClassNames(styles.input, styledClassName)}
+        />
+        <span className={styles.icon} onClick={handleShowPassword}>
```
pducolin, line 125: Is `span` accessible ? (in the a11y meaning)

pducolin:
Welcome to the `ui-components` jungle @a-coruble !!

docs/assets.json
```diff
@@ -1,29 +1,38 @@
 {
-  "vendors.js": "/ui-components/static/js/vendors.13ca2346.js",
```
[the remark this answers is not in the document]
a-coruble: Yes I am, but it's not the first time I have this issue at Dashlane, is there something I can do to rollback and remove this from the PR ?

src/components/password-input/PasswordInput.tsx
a-coruble: As the PasswordStrength will accept only a scoring between 0-4, the developer who uses this lib will be responsible for translating his own score to a scale on 0-4

src/components/password-input/PasswordInput.tsx
a-coruble: I added two callback functions, one triggered when you hide, the other when you reveal

src/components/password-input/PasswordInput.tsx
a-coruble: Okay, I removed the showPassword (this was bugging me too) and added 2 callback `onShowPassword` et `onHidePassword`

src/components/password-input/PasswordInput.tsx
a-coruble: I removed the default value and added a guard to display it only if not undefined. Otherwise it is not displayed at all

src/components/password-input/PasswordInput.tsx
a-coruble: Okay, those props have been marked as mandatory without default values

src/components/password-input/subcomponents/PasswordStrength/PasswordStrength.tsx
a-coruble: I agree, changed it

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
+import * as React from 'react';
+
+import { getClassNames } from '@utility/cssUtils';
+import { HideIcon, RevealIcon } from '@atoms/icons';
+import { midGreen00 } from '@colors';
+import { Tooltip } from 'components';
```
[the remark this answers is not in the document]
a-coruble: When I use `@components`, everything works great except when running the tests, I have this error : 

```
 FAIL  src/components/password-input/PasswordInput.spec.tsx
  ● Test suite failed to run

    Cannot find module '@components' from 'PasswordInput.tsx'

    However, Jest was able to find:
    	'./PasswordInput.mdx'
    	'./PasswordInput.module.scss'
    	'./PasswordInput.spec.tsx'
    	'./PasswordInput.tsx'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['js', 'json', 'jsx', 'ts', 'tsx', 'node'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string



      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:259:17)
      at Object.<anonymous> (src/components/password-input/PasswordInput.tsx:1266:26)```

File: src/components/password-input/PasswordInput.tsx [hunk 13]
```diff
@@ -0,0 +1,153 @@
...
+type _ForbiddenProps = 'size' | 'prefix' | 'type';
+
+export interface PasswordInputProps
+  extends Omit<React.InputHTMLAttributes<HTMLInputElement>, _ForbiddenProps> {
+  /**
+   * Some text to display under the input when triggering a feedback type
+   */
+  feedbackText?: string;
+  /**
+   * Some text to display under the input when triggering a feedback type
```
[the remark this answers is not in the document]
a-coruble: Good catch, indeed, changed it !

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,153 @@
...
+   * Props for the password strength indicator, see below for better description.
+   */
+  passwordStrengthProps?: PasswordStrengthProps;
+  /**
+   * Boolean triggering the display / hide of the password
+   * @default false
+   */
+  showPassword?: boolean;
+  /**
+   * Password strength indicator (score from 0 to 4)
```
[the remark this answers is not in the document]
a-coruble: This has been removed as it's now 2 separate components

src/components/password-input/PasswordInput.tsx
a-coruble: Now that you say it, I don't think it is. I'm wondering which element could replace it and have a good meaning

src/components/password-input/PasswordInput.tsx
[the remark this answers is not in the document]
oclero: I've fixed the imports for tests in my branch (PR is open: https://github.com/Dashlane/ui-components/pull/47) Please review it :)

docs/assets.json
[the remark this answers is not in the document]
pducolin: you can drop the commits called `docs: update docz documentation`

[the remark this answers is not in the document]
pducolin: these are artifacts, as we don't have CI on this repo we build it locally and store it on the repo. So not a big deal to have them malformatted

src/components/password-input/PasswordInput.tsx
pducolin: I think the only accessible elements are button and input, you could use a custom input of type checkbox here

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+    } else if (!showPassword && onHidePassword) {
+      onHidePassword();
+    }
+    setShowPassword(prevState => !prevState);
+  };
+
+  const styledClassName =
+    feedbackType && styles[feedbackType] ? styles[feedbackType] : '';
+
+  const Icon = showPassword ? HideIcon : RevealIcon;
```
pducolin, line 75: I would name this Icon to something more explicit, like `RevealOrHideIcon`

src/components/password-input/PasswordInput.tsx
a-coruble: Tried the `<input type='checkbox'>` but it does not accept a children, so I went with button and removed the native styling of a button.

src/components/password-input/PasswordInput.tsx
a-coruble: Indeed, changed it

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  hidePasswordTooltipText: string;
+  /**
+   * The label used to describe the field, it's displayed on top.
+   * Note that the text will be uppercase
+   */
+  label?: string;
+  /**
+   * A function that is triggered when the password gets revealed
+   */
+  onShowPassword?: () => void;
```
pducolin, line 37: why not a single callback for `onPasswordVisibilityChanged` ?

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
+  const {
+    feedbackText,
+    feedbackType,
+    hidePasswordTooltipText,
+    label,
+    onShowPassword,
+    onHidePassword,
+    showPasswordTooltipText,
+    ...rest
```
pducolin, line 58: I prefer `...prop` but might be only me

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <div className={styles.root}>
+      {label && <span className={styles.label}>{label}</span>}
+      <div className={styles.container}>
+        <input
+          {...rest}
+          type={showPassword ? 'text' : 'password'}
+          className={getClassNames(styles.input, styledClassName)}
+        />
+        <button className={styles.icon} onClick={handleShowPassword}>
```
pducolin, line 86: Yay!

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  const [showPassword, setShowPassword] = React.useState(false);
+
+  const handleShowPassword = () => {
+    if (showPassword && onShowPassword) {
+      onShowPassword();
+    } else if (!showPassword && onHidePassword) {
+      onHidePassword();
+    }
+    setShowPassword(prevState => !prevState);
```
pducolin, line 69: why not `setShowPassword(!showPassword)` ?

a-coruble: As stated in the documentation, the state updates can be run asynchronously in React, so I prefer to use the previous state to make sure the component has the expected behaviour.

src/components/password-input/PasswordInput.tsx
a-coruble: You're right, better syntax, I'll change it

src/components/password-input/PasswordInput.tsx
a-coruble: I do prefer having two function to get a good control on the callback executed, and as an external developer, I found it more logic to be able to have a separation of concern and capability in 2 interfaces.

src/components/password-input/PasswordInput.module.scss
```diff
@@ -0,0 +1,89 @@
...
+        background: colors.$dash-green-06;
+        cursor: not-allowed;
+      }
+
+      &.warning {
+        border: 1px solid colors.$orange-00;
+      }
+
+      &.error {
+        border: 1px solid colors.$functional-red-02;
```
oclero, line 76: If only the color changes, you could use border-color

src/components/password-input/PasswordInput.tsx
oclero: I don't agree. It would be like a checkbox having both `onChecked()` and `onUnchecked()`. For booleans, it's only two functions, but let's imagine an enum with 10 possible values. Would you do 10 functions? For each property I'm in favor of the standard get/set/notify trio of functions.

src/components/password-input/PasswordInput.tsx
oclero: What if there are 2 hide requests following each other? Will the second one actually *show* the password?

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+          {prop.disabled ? (
+            <RevealOrHideIcon disabled={prop.disabled} />
+          ) : (
+            <Tooltip
+              placement='left'
+              title={
+                showPassword ? hidePasswordTooltipText : showPasswordTooltipText
+              }
+            >
+              <RevealOrHideIcon hoverColor={midGreen00} />
```
oclero, line 96: I'm still incertain we should have colors hardcoded in ts files

src/components/password-input/PasswordInput.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (showPassword && onShowPassword) {
+      onShowPassword();
+    } else if (!showPassword && onHidePassword) {
+      onHidePassword();
+    }
+    setShowPassword(prevState => !prevState);
+  };
+
+  const styledClassName =
+    feedbackType && styles[feedbackType] ? styles[feedbackType] : '';
```
oclero, line 73: You can use `getClassNames` like this to move the classes at the same lace in the code:

```tsx
<span className={getClassNames(styles.feedback, {
  // Will add the class only if right part is true 
  [styles[feedbackType]] : feedbackType && styles[feedbackType]
})}>
  {feedbackText}
</span>
```

src/components/password-strength/PasswordStrength.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+import { getClassNames } from '@utility/cssUtils';
+
+import styles from './PasswordStrength.module.scss';
+
+export interface PasswordStrengthProps {
+  /**
+   * Any text you want to display under the strength indicator
+   */
+  additionalText?: string;
```
oclero, line 11: Is it up to the dev using the component to map score to text and update the text based on the score?

a-coruble: Yes, as we have no idea on how someone will use the correlation between the two. They might even be using the text zone to display something unrelated.

src/components/password-input/PasswordInput.tsx
a-coruble: Thanks ! Will change it !

src/components/password-input/PasswordInput.tsx
a-coruble: I'm divided on this one. On one hand, in cases like this one, it could be useful, in other, I find that we are splitting the display in both SCSS and TS which is not ideal as we decided not to go for a CSS-in-JS

src/components/password-input/PasswordInput.module.scss
a-coruble: Indeed, will change it

src/components/password-strength/PasswordStrength.tsx
oclero: OK I see. So it's up to the designers/copywriters to decide how this text behaves, and keep it consistent everywhere at Dashlane.

src/components/password-input/PasswordInput.tsx
oclero: Yes... Let's see later when we add dark theme :)

src/components/password-input/PasswordInput.tsx
a-coruble: The thing is that it has this behavior by scopes (https://www.freecodecamp.org/news/functional-setstate-is-the-future-of-react-374f30401b6b/) and batches calls inside one function. So that means that you'll never be able to make two hide requests, because it's going to execute the first, and re-render before you can click on the button again. Which leads to the expected behavior. But as the next state is computed on the previous state, it's better to use this syntax and avoid passing directly `true` or `false`.

src/components/password-input/PasswordInput.tsx
a-coruble: The thing here is that I might not have the feedbackType (`undefined` as not required), so TS won't let me use it as index in the object. Moreover, I have two elements which needs to apply this classname, so I prefer to set it once and forward it directly where needed. If this is really an issue for you, I'll change this behavior by adding a default value to the feedbackType, but I'd prefer not to

src/components/password-input/PasswordInput.tsx
a-coruble: After re-reading these comments, indeed I think you're right, I'll change those 2 functions into one `onPasswordVisibilityChanged`

src/components/password-input/PasswordInput.tsx
oclero: I see. I had kind of the same problem with Button's Nature, but FeedbackType is deffirent because `undefined` is allowed and makes sense (No feedback means no feedback (lol), but no Nature for a Button didn't make sense). It means we should have a `None` value for FeedbackType (we did things like this in C# on UWP but it's a different platform).  You can keep what you did :)

src/components/password-input/PasswordInput.tsx
oclero: Then perfect!

src/components/password-strength/PasswordStrength.tsx
a-coruble: Absolutely, and we are already doing that in the existing :)

PR #47, opened 2020-04-09
Feat/move button to components

oclero:
`Button` has been moved to components and refactored to use CSS instead of StyledComponents.


src/components/text-input/TextInput.tsx
```diff
@@ -8,31 +8,35 @@ type _ForbiddenProps = 'size' | 'prefix' | 'type';
...
-        {...props}
-        className={getClassNames(styles.input, styledClassName)}
+        {...htmlInputProps}
+        className={getClassNames(styles.input, feedbackCssClass)}
         type='text'
       />
-      {shouldShowFeedback(props) ? (
-        <span className={getClassNames(styles.feedback, styledClassName)}>
-          {props.feedbackText}
+      {shouldShowFeedback(feedbackText, feedbackType) ? (
```
ni0ki, line 37: At this point, it would be more straightforward to remove the shouldShowFeedback method and simply replace the call by `feedbackText && feedbackType`

src/components/button/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+  );
+};
+
+Button.defaultProps = {
+  nature: defaultNature,
+  size: defaultSize,
+  theme: defaultTheme
+};
+
+export default Button;
```
ni0ki, line 47: Do we need a named AND a default export ? I'd keep only the first one

oclero: Indeed I'm fixing this!

src/components/button/Button.tsx
```diff
@@ -0,0 +1,47 @@
...
+      {children ? <span className={styles.content}>{children}</span> : null}
+    </button>
+  );
+};
+
+Button.defaultProps = {
+  nature: defaultNature,
+  size: defaultSize,
+  theme: defaultTheme
+};
```
a-coruble, line 45: I think we can remove those as the `defaultProps` are used when we don't have Typescript. And in addition, they're already set to the default value at the beginning of the component.

src/components/index.ts
```diff
@@ -5,6 +5,7 @@ export { default as Tooltip } from './tooltip/Tooltip';
 export { TextInput } from './text-input/TextInput';
 export { Alert } from './alert/Alert';
 export { AlertSeverity } from './alert/types';
+export { default as Button } from './button/Button';
```
a-coruble, line 8: If you keep only the named export, make sure to change this line to 
```suggestion
export { Button } from './button/Button';
```

oclero: I keep only the default export, actually.

oclero: Edit: finally the named exports are preferred

src/components/button/Button.mdx
```diff
@@ -0,0 +1,137 @@
+---
+name: Button
+menu: Components
+---
+
+import { Playground, Props } from 'docz';
+import { Button } from './Button';
+import CopyPasswordIcon from '../../atoms/icons/CopyPasswordIcon';
```
pducolin, line 8: could use `@atoms/icons` and `@colors/colors`

src/components/button/Button.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { getClassNames } from '@utility/cssUtils';
+import styles from './Button.module.scss';
+
+type ButtonNature = 'primary' | 'secondary' | 'danger' | 'ghost';
```
maryiam, line 5: I think that it's better to keep the types that are not only used inside the component and that are exported (to clients or to other components in this lib) in a separate file `types.ts`.

src/components/button/Button.spec.tsx
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    it('should render danger nature and small size', () => {
+      const button = mount(<Button nature='danger' size='small' />);
+
+      expect(button).toMatchSnapshot();
+    });
+  });
+
+  describe('Children render', () => {
```
maryiam, line 59: I think we should test how the buttons behave for the different states (hover, active ..). These tests were present in `src/ions/button/themes` for each button. I think we should absolutely get them back :)

src/components/button/Button.tsx
```diff
@@ -0,0 +1,35 @@
...
+type ButtonSize = 'small' | 'medium' | 'large';
+type ButtonTheme = 'light' | 'dark';
+
+interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+  nature?: ButtonNature;
+  size?: ButtonSize;
+  theme?: ButtonTheme;
+}
+
+export const Button: React.FC<ButtonProps> = ({
```
maryiam, line 15: I love this ❤️ Very clean :)

src/components/button/Button.spec.tsx
pducolin: can we force the hover to take a snapshot?

src/components/button/Button.tsx
pducolin: Maybe in a subfolder then, so that we can prevent from exporting them out of the component folder (by preventing import from subfolders)

src/components/button/Button.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { getClassNames } from '@utility/cssUtils';
+import styles from './Button.module.scss';
+
+type ButtonNature = 'primary' | 'secondary' | 'danger' | 'ghost';
+type ButtonSize = 'small' | 'medium' | 'large';
+type ButtonTheme = 'light' | 'dark';
+
+interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
pducolin, line 9: Should we export this props ?

src/components/quick-actions-menu/QuickActionsMenu.mdx
```diff
@@ -4,7 +4,7 @@ menu: Components
 ---
 
 import { Playground, Props } from 'docz';
-import Button from '../../atoms/button/Button';
+import { Button } from '../../components/button/Button';
```
pducolin, line 7: Here you don't use `@component` to avoid circular import right?

src/components/text-input/TextInput.tsx
```diff
@@ -3,36 +3,32 @@ import { getClassNames } from '@utility/cssUtils';
...
+  const feedbackCssClass =
+    feedbackType && styles[feedbackType] ? styles[feedbackType] : '';
 
   return (
-    <div className={getClassNames(styles.root, styledClassName)}>
+    <div className={getClassNames(styles.root, feedbackCssClass)}>
       <input
-        {...props}
-        className={getClassNames(styles.input, styledClassName)}
+        {...htmlInputProps}
```
pducolin, line 25: love the name!

src/components/button/Button.spec.tsx
maryiam: I used to do this with `[jest-styled-components](https://github.com/styled-components/jest-styled-components)` like this
```ts
          const DangerButton = mount(<ButtonDanger />);
          const options = {
            modifier: ':hover:not(:disabled)'
          };
          expect(DangerButton).toHaveStyleRule(
            'background-color',
            functionalRed01,
            options
          );
```

Maybe we can find an alternative?

src/components/button/Button.tsx
maryiam: I'm not sure to understand. Why we should prevent exporting them?

src/components/button/Button.spec.tsx
oclero: I don't think we should test the style, but rather test that the UI interactions are not broken.

src/components/button/Button.tsx
oclero: Yes me too! The code has melt like ice, I love it! So small and compact :)

src/components/quick-actions-menu/QuickActionsMenu.mdx
oclero: I cannot import with aliases in Docz markdown files

src/components/button/Button.tsx
oclero: We can decide the convention for all components:
- Should we export props? (If so, we should name them `<Component>Props` to avoid confusing names. Ex: `ButtonProps`)
- Should we export the components' types like `ButtonSize`?

src/components/button/Button.tsx
oclero: I've answered on [another thread](https://github.com/Dashlane/ui-components/pull/47/files/42d4231d4b8cb719a4a361cd5b7b228b79a8ff0c#r408019867). I'm OK to export them. We just have to decide and respect the same convention for all components.

src/components/quick-actions-menu/QuickActionsMenu.mdx
oclero: Docz does not support aliases. Maybe it needs more config

oclero: Docz's does not support aliases for imports

maryiam: I think it does, as we do it [here](https://github.com/Dashlane/ui-components/blob/master/src/components/dropdown/DropdownMenu.mdx)

src/components/button/Button.spec.tsx
pducolin: If we could simulate the `hover` state and take a snapshot we could use that as a reference

src/components/quick-actions-menu/QuickActionsMenu.mdx
oclero: thanks, I was doing it the wrong way

PR #49, opened 2020-04-10
Feat/use named exports

oclero:
All module exports and imports now use "named" exports.


src/atoms/icons/index.ts
```diff
@@ -1,49 +1,49 @@
...
-export { default as CheckIcon } from './CheckIcon';
-export { default as CaretDownIcon } from './CaretDownIcon';
-export { default as LoaderIcon } from './LoaderIcon';
-export { default as SharedIcon } from './SharedIcon';
-export { default as NoNetworkIcon } from './NoNetworkIcon';
-export { default as HideIcon } from './HideIcon';
-export { default as RevealIcon } from './RevealIcon';
-export { default as CopyPasswordIcon } from './CopyPasswordIcon';
-export { default as EditIcon } from './EditIcon';
+export { LinkIcon } from './LinkIcon';
```
maryiam, line 2: You need to modify the following action in the `plopfile.js` to remove `default as` since the content of this auto generated :
```ts
{
        type: 'append',
        path: 'src/atoms/icons/index.ts',
        unique: true,
        pattern: /(\/\*-- EXPORT ICON COMPONENTS --\*\/)/gi,
        template:
          "export { {{pascalCase name}}Icon } from './{{pascalCase name}}Icon';"
      }
```

src/components/button/Button.mdx
```diff
@@ -0,0 +1,137 @@
+---
+name: Button
+menu: Components
+---
+
+import { Playground, Props } from 'docz';
+import { Button } from './Button';
+import { CopyPasswordIcon } from '../../atoms/icons/CopyPasswordIcon';
+import { white, midGreen00, dashGreen00 } from '../../colors/colors.ts';
```
a-coruble, line 9: Can we use the `@atoms, @components, @colors` instead of relative imports ?

config/generator/icons/IconComponentTemplate.tsx.hbs
```diff
@@ -1,14 +1,11 @@
 import * as React from 'react';
-import SvgWrapper from '@ions/icon/wrapper/SvgWrapper';
+import { SvgWrapper } from '@ions/icon/wrapper/SvgWrapper';
 
-const {{pascalCase name}}Icon: React.FunctionComponent<
-  import('../../ions/icon/types').IconProps
-> = (props: import('../../ions/icon/types').IconProps): JSX.Element => (
+export const {{pascalCase name}}Icon: React.FunctionComponent< import('../../ions/icon/types').IconProps> = (props:
+  import('../../ions/icon/types').IconProps): JSX.Element => (
```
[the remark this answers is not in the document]
oclero: It's not the scope and the goal of this PR, this why I didn't do it. But I agree we should do that

src/components/button/Button.mdx
oclero: Done in the Button's PR

PR #50, opened 2020-04-10
Feat: add toggle button component

akamkaro:


bradcush:
Why have 63 files changed as a result of this work?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+  labelOff?: string;
+  labelOn?: string;
+}
+
+const HiddenInput: React.FC<ToggleProps> = props => {
+  // const { checked, defaultChecked, labelOn, labelOff, ...inputProps } = props;
+  const { labelOn, labelOff, ...inputProps } = props;
+  return (
+    <input className={styles.hiddenInput} type='checkbox' {...inputProps} />
+    // defaultChecked={checked}
```
oclero, line 15: Can you remove these comments, if they're not useful ?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const Label: React.FC<ToggleProps> = props => {
+  return (props.checked && props.labelOn) ||
+    (!props.checked && props.labelOff) ? (
+    <span className={styles.text}>
+      {props.checked ? props.labelOn : props.labelOff}
+    </span>
+  ) : null;
+};
```
oclero, line 50: I think you can avoid giving full props here as argument, and instead just use an object containing only what you need:

```ts
const Label: React.FC<ToggleProps> = { checked: boolean, labelOn: string, labelOff: string } => {
  return (checked && labelOn) || (!checked && labelOff) ? (
    <span className={styles.text}>
      {checked ? labelOn : labelOff}
    </span>
  ) : null;
};
```

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+const Handle: React.FC<ToggleProps> = props => {
+  return (
+    <div
+      className={getClassNames(styles.handle, {
+        [styles.checked]: props.checked,
+        [styles.disabled]: props.disabled
+      })}
+    />
+  );
+};
```
oclero, line 41: Same here: just use what you need as function aguments: `{ checked: boolean, disabled: boolean }`

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+const HiddenInput: React.FC<ToggleProps> = props => {
+  // const { checked, defaultChecked, labelOn, labelOff, ...inputProps } = props;
+  const { labelOn, labelOff, ...inputProps } = props;
+  return (
+    <input className={styles.hiddenInput} type='checkbox' {...inputProps} />
+    // defaultChecked={checked}
+  );
+};
+
+const Groove: React.FC<ToggleProps> = props => {
```
oclero, line 19: Same remark here for the function argument

src/components/toggle/__snapshots__/Toggle.spec.tsx.snap
```diff
@@ -0,0 +1,202 @@
...
+<Component
+  checked={true}
+  defaultChecked={true}
+>
+  <label
+    className="root checked"
+  >
+    <HiddenInput
+      checked={true}
+      defaultChecked={true}
```
oclero, line 13: Do the tests really are up to date with what's in Toggl.ts? It doesn't look so.

oclero:
> Why have 63 files changed as a result of this work?

Because of the documentation that is unfortunately also part of this repository.

src/components/toggle/Toggle.tsx
akamkaro: Yes 👍

akamkaro:
> Why have 63 files changed as a result of this work?

Yes, as @oclero wrote it is because of the documentation :/

maryiam:
> Why have 63 files changed as a result of this work?

We don't have a CI for this project yet. So now, we need to do this to keep the doc https://dashlane.github.io/ui-components/ updated with the new changes.
We have a post-commit hook that looks if a developer changed something in the documentation in his commit that he has just created (ie changed files with `.mdx` extension). If so, we create a separate commit that includes the result of the command `yarn docz:build`

This is mentioned in the [CONTRIBUTING file](https://github.com/Dashlane/ui-components/blob/master/CONTRIBUTING.md#good-to-know-):
> We also execute the task yarn exec-docz-update on post-commit that will check if we have added or updated any Docz documentation (by modifying or adding any .mdx file), regenerate new built documentation and add push it through a new dedicated commit : in this way, we assure that https://dashlane.github.io/ui-components is always up to date.

src/components/toggle/__snapshots__/Toggle.spec.tsx.snap
pducolin: To explain about tests, since we test the Toggle.tsx and not ToggleDemo.tsx, setting `checked` value triggers an error, because `onChecked` is expected to be handled.  So we need to use the defaultChecked in the definition of the test.

pducolin:
@maryiam I'm thinking of creating a FAQ.md with this question, what do you think?

maryiam:
> @maryiam I'm thinking of creating a FAQ.md with this question, what do you think?

Oh god, that would be great !! 😃Thanks for proposing this ❤️

src/components/toggle/Toggle.mdx
```diff
@@ -0,0 +1,48 @@
...
+The example below shows an interactive `Toggle`.
+
+<Playground
+  style={{
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-around'
+  }}
+>
+  <ToggleDemo />
```
maryiam, line 47: Why did you need a demo? the Playground wasn’t enough to show all possible behaviors?

src/components/toggle/Toggle.module.scss
```diff
@@ -0,0 +1,140 @@
...
+        }
+      }
+    }
+  }
+}
+
+.groove {
+  display: flex;
+  align-items: center;
+
```
maryiam, line 114: Empty line

src/components/toggle/Toggle.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    it('should render disabled', () => {
+      const toggle = mount(<Toggle disabled />);
+      expect(toggle).toMatchSnapshot();
+    });
+
+    it('should render checked', () => {
+      const toggle = mount(<Toggle checked defaultChecked />);
+      expect(toggle.props().checked).toEqual(true);
```
maryiam, line 20: Is this test still relevant?  I don't see any prop `defaultChecked ` in the component. 

Also can we add a test to check that the input is **not checked** by default?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { getClassNames } from '@utility/cssUtils';
+import styles from './Toggle.module.scss';
+
+interface ToggleProps extends React.InputHTMLAttributes<HTMLInputElement> {
+  labelOff?: string;
+  labelOn?: string;
+}
+
+const HiddenInput: React.FC<ToggleProps> = props => {
+  // const { checked, defaultChecked, labelOn, labelOff, ...inputProps } = props;
```
maryiam, line 11: comment to remove

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+      className={getClassNames(styles.handle, {
+        [styles.checked]: props.checked,
+        [styles.disabled]: props.disabled
+      })}
+    />
+  );
+};
+
+const Label: React.FC<ToggleProps> = props => {
+  return (props.checked && props.labelOn) ||
```
maryiam, line 44: I don't really understand these conditions. What do `labelOn` and `labelOff ` refer to? 
For better readability, can you please declare each condition in a `const` and give it an explicit name?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+import styles from './Toggle.module.scss';
+
+interface ToggleProps extends React.InputHTMLAttributes<HTMLInputElement> {
+  labelOff?: string;
+  labelOn?: string;
+}
+
+const HiddenInput: React.FC<ToggleProps> = props => {
+  // const { checked, defaultChecked, labelOn, labelOff, ...inputProps } = props;
+  const { labelOn, labelOff, ...inputProps } = props;
```
maryiam, line 12: If you don't need `labelOn` and `labelOff` then maybe don't pass them to `HiddenInput` ?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+        [styles.checked]: props.checked,
+        [styles.disabled]: props.disabled
+      })}
+    >
+      {props.children}
+    </div>
+  );
+};
+
+const Handle: React.FC<ToggleProps> = props => {
```
maryiam, line 32: I don't this that the name `Handle` reflects what this component is supposed to do? What should it be handling?

src/components/toggle/__snapshots__/Toggle.spec.tsx.snap
maryiam: I don't understand? You mean that you modify manually the content of `Toggle.spec.tsx.snap` ?

src/components/toggle/Toggle.tsx
maryiam, line 19: What is `Groove`?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+  // const { checked, defaultChecked, labelOn, labelOff, ...inputProps } = props;
+  const { labelOn, labelOff, ...inputProps } = props;
+  return (
+    <input className={styles.hiddenInput} type='checkbox' {...inputProps} />
+    // defaultChecked={checked}
+  );
+};
+
+const Groove: React.FC<ToggleProps> = props => {
+  return (
```
maryiam, line 20: This applies for all components in this file, you don't need the 

```ts
return (
...
)
```
since you have only one instruction in your function, you can 
```ts
const Groove: React.FC<ToggleProps> = ({ checked: boolean, disabled: boolean }) => (<div>...</div)
```

src/components/toggle/__snapshots__/Toggle.spec.tsx.snap
maryiam: The `defaultChecked` prop no longer exist in the component.

src/components/toggle/Toggle.tsx
oclero: And we could rename `labelOn` and `labelOff` as `labelChecked` and `labelUnchecked`, for naming consistency with `checked` and `unchecked`

src/components/toggle/Toggle.tsx
oclero: It's basically the same names in a slider (groove and handle), for me it's OK to use it.

src/components/toggle/Toggle.tsx
akamkaro: Not possible. If i want to pass only the 3 arguments i need, i have to create another interface.
In my opinion, the best solution here is to do like this :
`const Label: React.FC<ToggleProps> = ({ checked, labelOn, labelOff }) => {
  return (checked && labelOn) || (!checked && labelOff) ? (
    <span className={styles.text}>{checked ? labelOn : labelOff}</span>
  ) : null;
};`

`<Label {...props} />`

src/components/toggle/Toggle.mdx
akamkaro: With the playground you have zero interaction, because the isChecked state is outside the Toggle component.
So we created a ToggleDemo for the isChecked state :
`export const ToggleDemo = () => {
  const [isChecked, setIsChecked] = React.useState(false);
  return (
    <div>
      <Toggle
        labelOn='Activated'
        labelOff='Deactivated'
        checked={isChecked}
        onChange={e => setIsChecked(e.currentTarget.checked)}
      />
    </div>
  );
};`

src/components/toggle/Toggle.tsx
akamkaro: labelOn (now labelChecked) is the label displayed when the toggl is checked.
labelOff (now labelUnchecked) is the label displayed when the toggl is unchecked.

I renamed these two properties as Olivier requested.

src/components/toggle/Toggle.tsx
akamkaro: The subComponent **Handle** is the grip (= poignée). 
The subComponent **Groove** is the background shape.

I agree with you it's not 100% explicit but I didn't find best names for these (these names are used in many toggle tutorials).

src/components/toggle/Toggle.spec.tsx
akamkaro: I removed the defaultChecked prop, it is now useless because i added a readonly prop in the tests to avoid a javascript alert about the onChange event.
I keep the "checked" test, and as you asked I added an unchecked test.

src/components/toggle/__snapshots__/Toggle.spec.tsx.snap
akamkaro: I removed it and added a "readonly" prop in the tests (cf other comment)

src/components/toggle/Toggle.mdx
```diff
@@ -0,0 +1,48 @@
...
+
+## Properties
+
+<Props of={Toggle} />
+
+## Usage
+
+### Basic usage
+
+Examples below shows the `Toggle` in its checked and unchecked states, both in enabled and disabled modes:
```
oclero, line 22: Small typo: it's "show", not "shows"

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+  labelChecked?: string;
+}
+
+const HiddenInput: React.FC<
+  React.InputHTMLAttributes<HTMLInputElement>
+> = props => (
+  <input className={styles.hiddenInput} type='checkbox' {...props} />
+);
+
+const Groove: React.FC<GrooveProps> = ({ checked, disabled, children }) => (
```
pducolin, line 32: just a remark and question point for guidelines: do we prefer
a) `const ComponentName: React.FC<Props> = ({prop1, prop2, ...props}) =>`
OR
b)  `const ComponentName = ({prop1, prop2, ...props}: Props) =>`

I would say `b` (more compact, same or better readable), what do you think @maryiam @bradcush ?

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+    })}
+  />
+);
+
+const Label: React.FC<LabelProps> = ({
+  checked,
+  labelChecked,
+  labelUnchecked
+}) => {
+  const isLabelCheckedVisible = checked && labelChecked;
```
pducolin, line 57: would refactor to `const isLabelVisible = (checked && labelChecked) || (!checked && labelUnchecked)`

src/components/toggle/Toggle.tsx
akamkaro: Cf other comment where i explained :)

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  return isLabelCheckedVisible || isLabelUncheckedVisible ? (
+    <span className={styles.text}>
+      {checked ? labelChecked : labelUnchecked}
+    </span>
+  ) : null;
+};
+
+export const Toggle: React.FC<ToggleProps> = props => {
+  const { labelChecked, labelUnchecked, ...hiddenInputProps } = props;
```
pducolin, line 68: would refactor to `export const Toggle: React.FC<ToggleProps> = { labelChecked, labelUnchecked, ...hiddenInputProps } `

src/components/toggle/Toggle.module.scss
```diff
@@ -0,0 +1,139 @@
+@use '../../colors/colors';
+@use '../../typography/typography';
+
+.root {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+
```
oclero, line 8: Can you remove this empty line?

src/components/toggle/Toggle.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    it('should render with labelOff', () => {
+      const labelUnchecked = 'Off';
+      const toggle = mount(<Toggle labelUnchecked={labelUnchecked} />);
+      expect(toggle.find('span').text()).toEqual(labelUnchecked);
+      expect(toggle).toMatchSnapshot();
+    });
+  });
+});
```
oclero, line 44: Here is a test we can add: checking if the callback is still called when clicking on the component

src/components/toggle/Toggle.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    it('should render checked', () => {
+      const toggle = mount(<Toggle checked />);
+      expect(toggle.props().checked).toEqual(true);
+      expect(toggle).toMatchSnapshot();
+    });
+
+    it('should render unchecked', () => {
+      const toggle = mount(<Toggle />);
+      expect(toggle.props().checked).toEqual(false || undefined);
```
maryiam, line 26: why isn’t it always `undefined` ?

src/components/toggle/Toggle.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+      expect(toggle).toMatchSnapshot();
+    });
+
+    it('should render with labelOn', () => {
```
maryiam, line 30: I think that you should rename `labelOn` and `labelOff` in the tests as well to reflect the change.

src/components/toggle/Toggle.spec.tsx
maryiam: It would be great indeed to test if the component behaves as expected on click!

src/components/toggle/Toggle.tsx
maryiam: I personally use both. I think that it's more of a team decision and it's a good idea that you started the conversation in the channel :)

src/components/toggle/Toggle.tsx
```diff
@@ -0,0 +1,88 @@
...
+        [styles.checked]: props.checked,
+        [styles.disabled]: props.disabled
+      })}
+    >
+      <HiddenInput {...hiddenInputProps} />
+      <Groove checked={props.checked} disabled={props.disabled}>
+        <Handle checked={props.checked} disabled={props.disabled} />
+      </Groove>
+      <Label
+        checked={props.checked}
```
maryiam, line 82: For consistency, I would rather destruct `checked` as well and avoid having the mix of `props.something` and destructed props (like `labelUnchecked `)

oclero: The destructuring is made to ensure the HTML `<input>`  won't receive props it doesn't need (there is a warning in console if we don't do that)

src/components/toggle/Toggle.spec.tsx
akamkaro: You're right :)

src/components/toggle/Toggle.tsx
akamkaro: As it is not a team guideline, i won't change this for the moment.

akamkaro:
>     * Small typo (quick to fix)

What do you mean by "small typo" ?

oclero:
> > ```
> > * Small typo (quick to fix)
> > ```
> 
> @oclero What do you mean by "small typo" ?
You've already fixed it ("show" instead of "shows").

src/components/toggle/Toggle.spec.tsx
akamkaro: As discussed with @maryiam , i'll implement more tests in an upcoming PR.

PR #51, opened 2020-04-14
feat: add explanations about dev usage in another project

ni0ki:
Small reminder that to use a local development version of ui-components within another project,
it might be necessary to link the react depdency

### DIFF
--- a/documentation/EXTERNAL_PROJECT_DEV_INTEGRATION.md
+++ b/documentation/EXTERNAL_PROJECT_DEV_INTEGRATION.md
@@ -1,0 +1,32 @@
+# Use local dev version inside your project
+
+When you want to try some changes you've made on a local version in your project, there's an extra step required to ensure there won't be any React conflict between the 2 projects.
+
+## Use the react dependency of your project
+
+To build _ui-components_ with another react version, you need to link it using NPM **after** installing it :
+
+```
+# Inside ui-components:
+
+# Install the dependencies
+yarn
+
+# Use the react version of the other project
+npm link $PATH_TO_FINAL_PROJECT/node_modules/react
+
+# Build ui-components using the linked react version
+yarn build
+```
+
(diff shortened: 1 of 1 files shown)


documentation/EXTERNAL_PROJECT_DEV_INTEGRATION.md
```diff
@@ -0,0 +1,32 @@
+# Use local dev version inside your project
```
maryiam, line 1: I don't understand the need of this file, we already describe this here https://github.com/Dashlane/ui-components/blob/master/CONTRIBUTING.md#developing-in-your-project

ni0ki: I did not see someone else already wrote it. Sorry.

oclero: Should we move documentation to the documentation/ folder, in separate files. IMO it's better to have small files dedicated to ony one subject, instead of one gigantic CONTRIBUTING.md file (that nobody reads). Fortnately, ours is not very long (for the moment). We'll see later if we need that.

ni0ki: As stated by Maryiam above, this documentation is not needed. I'm closing the PR.
If we want to split the existing CONTRIBUTING.md, let's do it in another PR.

pducolin: I agree with Olivier, IMHO having smaller files could improve readability of the whole documentation. I would refactor the `CONTRIBUTING.md` removing this part and adding a link to this guide instead

oclero: Contributing file is still small, so it's not urgent to refactor it.

PR #52, opened 2020-04-14
docs: add FAQ page

pducolin:
This is just the beginning of it, with the most asked question.


oclero:
> the most asked question.

What's the meaning of life and the Universe?

pducolin:
42

PR #54, opened 2020-04-15
Feature/update docz to 2.3

ni0ki:
DocZ 2.3 finally landed a few days ago.
This PR updates _ui-components_ with DocZ 2.3.1

new features:

- Uses Gatsby as generator
- Ability to hide default HTML props for some components
- Theming now supported (but not used in this PR)
- Ability to  shadow / override any components used by DocZ

Small warning:
It's not as easy to style the Playground as before. I had to override the base component to support passing a custom style object per playground.
(More on that : [DocZ Component Shadowing](https://www.docz.site/docs/component-shadowing)

For now only the preview part is customizable, but we can always improve that later on.

Example :
`<Playground style={{ preview: { backgroundColor: 'blue'} }}`


.eslintrc.json
```diff
@@ -69,6 +69,13 @@
       "rules": {
         "@typescript-eslint/no-explicit-any": "off"
       }
+    },
+    {
+      "files": ["src/gatsby-theme-docz/**/*.js"],
```
ni0ki, line 74: Files in this folder will only be DocZ component customization / rewrite.
As some of their content may copy or extend docz source, we don't want to be as strict as for our exported components.

package.json
```diff
@@ -87,11 +88,12 @@
     "sass-loader": "^8.0.2",
     "semantic-release": "^15.13.27",
     "ts-jest": "^24.0.2",
-    "typescript": "^3.5.3"
+    "typescript": "3.7.3"
   },
   "resolutions": {
     "webpack": "4.30.0",
-    "react-docgen-typescript": ">=1.13.0"
+    "react-docgen-typescript": ">=1.13.0",
```
ni0ki, line 95: This typescript version seems necessary to support native props filtering.
More here : [https://github.com/doczjs/docz/issues/337](https://github.com/doczjs/docz/issues/337)

package.json
```diff
@@ -87,11 +88,12 @@
     "sass-loader": "^8.0.2",
     "semantic-release": "^15.13.27",
     "ts-jest": "^24.0.2",
-    "typescript": "^3.5.3"
+    "typescript": "3.7.3"
```
oclero, line 91: Why not `^3.7.3` (instead of being stuck to a specific version)?

package.json
```diff
@@ -87,11 +88,12 @@
...
   },
   "resolutions": {
     "webpack": "4.30.0",
-    "react-docgen-typescript": ">=1.13.0"
+    "react-docgen-typescript": ">=1.13.0",
+    "typescript": "3.7.3"
```
oclero, line 96: Same question here

package.json
ni0ki: Sorry, I misplaced my comment. Quoting the original one :

> This typescript version seems necessary to support native props filtering.
More here : doczjs/docz#337

maryiam:
Hey, can you make a small demo (video maybe) showing how does the documentation looks like now? :)

.eslintrc.json
maryiam: Why is this needed? The sort import role should not be an issue and should be respected even if we are writing documentation. Also, we don't use `react/prop-types` with typescript, so I don't understand turning off this rule.

package.json
maryiam: Then let's keep this version in the `resolution` section rather than fixing the version here

ni0ki:
Small video to showcase the difference : https://files.dashlane.com/s/2Yx27Wf2Gw2m9kb

doczrc.js
```diff
@@ -1,30 +1,15 @@
 const TsConfigPathsPlugin = require('awesome-typescript-loader')
```
maryiam, line 1: This import should be removed then?

package.json
```diff
@@ -75,6 +74,8 @@
     "plop": "^2.5.0",
     "prettier": "^1.18.2",
     "pretty-quick": "^1.11.1",
+    "react": "^16.12.0",
```
maryiam, line 77: Why do we need to install `react` now? It's already in the  `peerDependencies` section.

src/atoms/icons/Icons.mdx
```diff
@@ -69,13 +69,14 @@ playground of the icons and add some props.
...
 <Playground
   style={{
-    backgroundColor: '#F5F6F7',
-    padding: '30px',
-    borderRadius: '4px',
-    display: 'flex',
-    flexDirection: 'row',
-    flexWrap: 'wrap',
-    justifyContent: 'space-around'
+    preview: {
```
maryiam, line 72: Is it possible to configure these styles with css/scss classes?

src/utility/cssUtils.ts
```diff
@@ -12,16 +12,20 @@ interface CssClasses {
...
 export const getClassNames = (...classes: (string | CssClasses)[]): string => {
   const result: string[] = [];
-  classes.forEach(value => {
-    if (typeof value === 'string') {
-      result.push(value);
-    } else {
-      Object.entries(value).forEach(([key, value]) => {
-        if (value) {
-          result.push(key);
+  if (classes) {
```
maryiam, line 15: Why do you need this?  I think that with the rest param `...classes` you init the value as an empty array even if you call `getClassNames` without params.

oclero: You're right, too many security checks!

maryiam:
> Small video to showcase the difference : https://files.dashlane.com/s/2Yx27Wf2Gw2m9kb

This is awsome @ni0ki  ❤️ Thanks :)

package.json
ni0ki: react & react-dom seem to be direct dependencies of docZ 2.x : https://www.docz.site/docs/getting-started

src/atoms/icons/Icons.mdx
ni0ki: Having a look on https://github.com/doczjs/docz/blob/master/core/docz/src/components/Playground.tsx, it seems we can use the className property. I'll check it out

doczrc.js
ni0ki: :+1:

.eslintrc.json
ni0ki: Agreed for the sort-import. I wanted to keep my playground code as close as the original one, but changing the imports order won't hurt.

About react/prop-types, it was already used by the project : [https://github.com/Dashlane/ui-components/blob/0513b6db7ec399709c856e45454dc3af7c38907c/.eslintrc.json#L64](https://github.com/Dashlane/ui-components/blob/0513b6db7ec399709c856e45454dc3af7c38907c/.eslintrc.json#L64)

package.json
ni0ki: Hmm, for unknown reasons, I need to keep both of them to 3.7.3 to have only the components props displayed (avoid having the HTMLInputProps listed).

.eslintrc.json
maryiam: Ok then let's add the docz files in this same section:
```js
    {
      "files": ["**/*.tsx", "src/gatsby-theme-docz/**/*.js""],
      "rules": {
        "react/prop-types": "off"
      }
    },
```

package.json
maryiam: Can you please make sure that the final build bundle doesn’t include them?

maryiam: As we always want the parent project to install react as dependency to avoid ending up with two react bundles in the final build of the parent project.

src/gatsby-theme-docz/components/Playground/index.js
```diff
@@ -0,0 +1,122 @@
+/**
+ * Original source : https://github.com/doczjs/docz/blob/master/core/gatsby-theme-docz/src/components/Playground/index.js
```
maryiam, line 2: So what are we overring here?

package.json
maryiam: Can you try with (while keeping the version in `devDependencies` as `^3.7.3`)
```json
  "resolutions": {
    "**/typescript": "3.7.3"
  },
```

.eslintrc.json
ni0ki: :+1:

doczrc.js
ni0ki: I just had a closer look inside the yarn.lock & dependencies. It seems we can't use ^3.7.3 as a version as it will resolve to 3.8.3 which is the latest TS version.
However, I can use ~3.7.3 which is compatible and ensure we still can benefit from any patch for this version. That's the best tradeoff I think

package.json
ni0ki: Unfortunately, it doesn't work either.
But... I just had a closer look inside the yarn.lock & dependencies. It seems we can't use ^3.7.3 as a version as it will resolve to 3.8.3 which is the latest TS version.
However, I can use ~3.7.3 which is compatible and ensure we still can benefit from any patch for this version. That's the best tradeoff I think

maryiam: Ok, I think that we can do this way:

1. remove typescript: `yarn remove typescript`
2. install typescript latest version `yarn add typescript`
3. fix resolution to the latest version of typescript
```json
  "resolutions": {
    "typescript": ">=3.8.3"
  },
```
In this way, we don't force to use an old version of typescript in the project.

ni0ki: It works. Beautiful \o/

src/gatsby-theme-docz/components/Playground/index.js
ni0ki: We make use of the style & className properties, so that anyone can easily customize the Playground for its need.

maryiam: I'm not able to find the Docz documentation regarding this. Can you please send a link to help me better understand this?

package.json
ni0ki: This should not conflict, as using ui-components as a lib inside another project won't install the react listed in devDependencies. This is only required by DocZ & Gatsby. There should not be any conflict.

src/components/dropdown/DropdownMenu.mdx
```diff
@@ -26,17 +25,17 @@ The dropdown can be defined alongside a controller. To display it correctly:
...
+    flexDirection: 'row',
+    alignItems: 'flex-start',
+    justifyContent: 'space-around',
+    height: '200px'
   }}
 >
   {() => {
   const [dropdownIsOpen, setDropdownIsOpen] = React.useState(false);
-  const controllerRef = React.useRef(null);
+    const controllerRef = React.useRef(null);
```
oclero, line 38: why this space? :)

PR #55, opened 2020-04-16
Feat/Accordion-Component

sanatshah:
PR to add in an accordion component!

There are 2 parts to it. Accordion and Accordion Sections. 

Accordion.Sections are the main components that define each Accordion row. You can wrap Accordion.Sections in an Accordion to have the Accordion control the open/close states of all the sections. Although it is not necessary to wrap Sections. 

Screen Shot 2020-04-16 at 11 08 25 AM

More info and examples in the documentation =)


sericaia:
Maybe you can add a snapshot in the PR too? :) can you also please rebase with master?

src/atoms/icons/ArrowIcon.tsx
```diff
@@ -0,0 +1,19 @@
...
+const ArrowIcon: React.FunctionComponent<
+  import('../../ions/icon/types').IconProps
+> = (props: import('../../ions/icon/types').IconProps): JSX.Element => (
+  <SvgWrapper {...props}>
+    <React.Fragment>
+      <path
+        fillRule='evenodd'
+        clipRule='evenodd'
+        d='M8 7.01567L1.55319 -9.53674e-07L0 1.29409L8 10L16 1.29409L14.4468 9.53674e-07L8 7.01567Z'
+        fill='#0E353D'
```
sericaia, line 13: should not have `fill`. Have you crated this Icon using the generator? check https://github.com/Dashlane/ui-components/blob/master/documentation/ICON_STEPBYSTEP.md

src/components/accordion/Accordion.module.scss
```diff
@@ -0,0 +1,33 @@
...
+  display: flex;
+  flex-direction: column;
+}
+
+.AccordionSectionHeader {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  min-height: 83px;
```
sericaia, line 15: 83px is a strange number

src/components/accordion/Accordion.spec.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
+import Accordion from './Accordion';
+
+describe.only('<Accordion />', () => {
```
sericaia, line 5: `.only` will only trigger this test, please remove it ;)

src/components/accordion/Accordion.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+    it('should be controlled, allowing only one section opened at a time', () => {
+      const accordionSection = mount(
+        <Accordion maxOneSectionOpen open={3}>
+          <Accordion.Section label='Some title' open>
+            Contet
+          </Accordion.Section>
+          <Accordion.Section label='Some title' open>
+            Contet
+          </Accordion.Section>
+          <Accordion.Section label='Some title'>Contet</Accordion.Section>
```
sericaia, line 72: s/contet/content

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+    }
+  }
+
+  const [opened, setOpened] = React.useState(initialOpenState);
+
+  const state = {
+    open: opened,
+    maxOneSectionOpen: props.maxOneSectionOpen,
+    allSectionsOpened: props.allSectionsOpened,
+    oneSectionAlwaysOpen: props.oneSectionAlwaysOpen,
```
sericaia, line 116: maybe you can use destruturing here

`const Accordion = ({maxOneSectionOpen, ... }: AccordionProps) => {`

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const Accordion = (props: AccordionProps) => {
+  let initialOpenState = props.open ? props.open : 0;
+  if (props.maxOneSectionOpen) {
+    if (Array.isArray(props.open)) {
+      initialOpenState = props.open[0];
+    }
+  }
+
+  const [opened, setOpened] = React.useState(initialOpenState);
```
sericaia, line 110: Not sure if it is inferred, if not add `React.useState<boolean>`

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+  return (
+    <AccordionContext.Provider value={state}>
+      <div className={ContainerClass}>{props.children}</div>
+    </AccordionContext.Provider>
+  );
+};
+
+Accordion.Section = Section;
+Accordion.Title = AccordionTitle;
+export default Accordion;
```
sericaia, line 134: don't use a default export :pray:

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+    return requestedIndex === open;
+  }
+};
+
+interface AccordionContext extends AccordionState {
+  childIndex: number;
+  onClick: (index: number) => void;
+}
+
+const AccordionContext = React.createContext(null as (AccordionContext | null));
```
sericaia, line 29: I would like to understand the reasoning to use context here, can you maybe add some comments about it

sericaia:
your commits are unverified, can you add a gpg key please?

src/components/accordion/Accordion.tsx
sericaia: btw this `(null as (AccordionContext | null));` seems crazy. Afaik `createContext` first param is the default value, why are you trying to add `AccordionContext`?

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+const useAccordionContext = () => {
+  const context = React.useContext(AccordionContext);
+  if (!context) {
+    throw new Error(
```
sericaia, line 34: thanks for doing this validation! 👍 💯

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const state = {
+    open: opened,
+    maxOneSectionOpen: props.maxOneSectionOpen,
+    allSectionsOpened: props.allSectionsOpened,
+    oneSectionAlwaysOpen: props.oneSectionAlwaysOpen,
+    childIndex: 0,
+    onClick: setOpened
+  };
+
+  const ContainerClass = props.accordionClassName
```
sericaia, line 121: why not concatenating both classes if there is a prop?

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,134 @@
...
+    ? props.accordionClassName
+    : styles.AccordionContainer;
+
+  return (
+    <AccordionContext.Provider value={state}>
+      <div className={ContainerClass}>{props.children}</div>
```
sericaia, line 127: can `ContainerClass` be camelCase?

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  sectionContentClassName?: string;
+
+  sectionContainerStyle?: {
+    backgroundColor?: string;
+    lockedOpacity?: string;
+  };
+}
+
+const ContainerDiv = styled.div<AccordionSectionProps>`
```
sericaia, line 25: we should not use styled components (unfortunately IMO!)

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,123 @@
...
+    setIsOpen(!isOpen);
+  };
+
+  let Header = (
+    <div className={styles.AccordionSectionHeader} onClick={onHeaderClick}>
+      <div>{label}</div>
+      <div style={{ float: 'right' }}>
+        <ArrowDownIcon
+          height={50}
+          width={27}
```
sericaia, line 70: usually icons are 20px or 30px, can you confirm this value?

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,123 @@
...
+        child
+          ? ((child as React.ReactElement).type as React.ComponentClass)
+              .displayName === 'AccordionTitle'
+          : false
+      )
+    : undefined;
+
+  let childrenToRender = children;
+
+  if (CustomHeaderIndex != undefined) {
```
sericaia, line 91: can't you simply use `!CustomHeaderIndex`?

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,123 @@
...
+          iconAnimated={false}
+          color={dashGreen00}
+          rotate={isOpen ? 180 : 0}
+        />
+      </div>
+    </div>
+  );
+
+  //See if there is a custom Header given
+  const CustomHeaderIndex = Array.isArray(children)
```
sericaia, line 80: I wonder if there isnt a better way to do this

sericaia: why the Header is not a prop? can you give an example of this in the docs?

src/components/accordion/__snapshots__/Accordion.spec.tsx.snap
```diff
@@ -0,0 +1,2707 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<Accordion /> Accordion should be controlled, allowing only one section opened at a time 1`] = `
+<Accordion
```
sericaia, line 4: These snapshots are long and it makes them not that useful. In the case I would prefer having another type of testing.

src/design-tokens/icons/icons-defs.json
```diff
@@ -46,5 +46,6 @@
...
+  "link": "<path d=\"M12.7562 9.34846L15.8049 12.4147C16.7362 13.3513 16.7334 14.8672 15.7986 15.8004C14.8639 16.7336 13.3511 16.7308 12.4198 15.7941L10.8212 14.1863L9.75988 15.2464L11.3582 16.8539C12.8737 18.378 15.3352 18.3826 16.8562 16.8641C18.3773 15.3456 18.3819 12.879 16.8664 11.3549L13.8177 8.28869C12.3023 6.76456 9.84076 6.76001 8.31972 8.27851C8.14747 8.45047 7.99467 8.63458 7.86133 8.82811L8.95176 9.91727C9.06123 9.71035 9.20309 9.51616 9.37733 9.3422C10.3121 8.409 11.8249 8.4118 12.7562 9.34846Z\" /><path d=\"M7.24385 10.6515L4.19513 7.58529C3.26382 6.64863 3.26663 5.1328 4.2014 4.1996C5.13617 3.26639 6.64893 3.26919 7.58024 4.20585L9.145 5.7796L10.2048 4.71804L8.64179 3.14608C7.12637 1.62196 4.66483 1.6174 3.14378 3.1359C1.62273 4.65441 1.61817 7.12094 3.13358 8.64506L6.1823 11.7113C7.69772 13.2354 10.1593 13.24 11.6803 11.7215C11.8496 11.5525 12 11.3718 12.1317 11.1819L11.0424 10.0938C10.9337 10.2966 10.7938 10.4869 10.6227 10.6578C9.68792 11.591 8.17516 11.5882 7.24385 10.6515Z\" />",
+  "arrow": "<path fill-rule=\"evenodd\" clip-rule=\"evenodd\" d=\"M8 7.01567L1.55319 -9.53674e-07L0 1.29409L8 10L16 1.29409L14.4468 9.53674e-07L8 7.01567Z\" fill=\"#0E353D\"/>"
```
sericaia, line 50: remove `fill`

src/ions/icon/base/Icon.tsx
```diff
@@ -3,13 +3,30 @@ import { Props as IconProps } from '@ions/icon/types';
...
+
+const getTransition = ({ svgAnimated, iconAnimated }: Props) => {
+  let transition;
+
+  if (svgAnimated) transition = 'none';
+  else transition = 'all 0.1s ease-in';
+
+  if (typeof iconAnimated !== 'undefined') {
+    //check if the user actually passed a iconAniamted prop
+    if (iconAnimated) transition = 'all 0.1s ease-in';
```
sericaia, line 24: can you use curly braces here?

src/ions/icon/base/Icon.tsx
```diff
@@ -3,13 +3,30 @@ import { Props as IconProps } from '@ions/icon/types';
 
 export type ColorProps = 'color' | 'hoverColor' | 'activeColor';
 interface Props extends Omit<IconProps, 'title' | 'children'> {
-  animated?: boolean;
+  svgAnimated?: boolean;
```
sericaia, line 6: I'm not sure if this is a breaking change or not

src/components/accordion/Section.tsx
sanatshah: Interesting, whats the case for not using styled components?

src/atoms/icons/ArrowIcon.tsx
sanatshah: Forgot to remove that field. Ended up using a icon that already existed anyways, so I removed the icon.

src/components/accordion/Accordion.module.scss
sanatshah: haha, It is a bit odd. Updated it to 80 px for more sanity.

src/components/accordion/Accordion.tsx
sanatshah: Good idea, thanks!

src/components/accordion/Accordion.tsx
sanatshah: Added some comments. I used context here so if the Accordion.Section is wrapped in an Accordion, the Accordion component can provide the context to its children(the open states). Some more info here : [component components with react hooks](https://kentcdodds.com/blog/compound-components-with-react-hooks). Had to add that to get the type checking to work properly when the state is later added as a value in the Accordion component.

src/components/accordion/Accordion.tsx
sanatshah: Updated!

src/components/accordion/Section.tsx
sanatshah: Updated to 30px!

src/components/accordion/Section.tsx
sanatshah: Typescript complains and wanted a explicit undefined check

src/components/accordion/Section.tsx
sanatshah: I think I have overthought this one. Was originally thinking that the experience of using a separate component might be better, but yes, after a revisit render prop makes a bit more sense. Plus the codes a bit cleaner

src/ions/icon/base/Icon.tsx
sanatshah: Should not be a breaking change. Just renamed the prop that gets used by the svgWrapper in order to make it more clear for where the animated config is coming from and support adding a iconAnimated property that can be used with the Icon Components.

src/design-tokens/icons/icons-defs.json
sanatshah: Ended up removing that icon!

src/components/accordion/Accordion.tsx
```diff
@@ -126,7 +126,7 @@ const Accordion = ({
   };
 
   const containerClass = accordionClassName
-    ? accordionClassName
+    ? `${styles.AccordionContainer} ${accordionClassName}`
```
oclero, line 129: I've added a small utility function to get the class names. You can use it like this:

```tsx
import { getClassNames } from '@utility/cssUtils';

const Comp = () => <component className={getClassNames(
  // You can use strings
  className1,                         // Always present
  // Or objects if you want to control if a class name is added or not
  {
    className2: true | false,         // Will be present if right side is evaluated to true
    [dynamicClassName]: true | false, // You can also use dynamic class names
  }
)}/>;
```

src/components/accordion/Section.tsx
sericaia: (unfortunately) this project uses sass

src/components/accordion/Section.tsx
sericaia: then cant you use !== instead of != ?

src/components/accordion/Section.tsx
oclero: It was decided to use pure CSS at runtime (Sass is only a build-time dependency)

src/components/accordion/Accordion.tsx
sanatshah: Thanks, updated!

src/components/accordion/__snapshots__/Accordion.spec.tsx.snap
sanatshah: Updated! Added enzyme/jest tests for the open/close behavior to programmatically determine if it is rendered appropriately instead of relying on snapshots

src/components/accordion/Section.tsx
sanatshah: Updated to not use styled components!

src/components/accordion/Accordion.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Accordion.Section label='Some title' open>
+          Contet
+        </Accordion.Section>
+        <Accordion.Section label='Some title' open>
+          Contet
+        </Accordion.Section>
+        <Accordion.Section label='Some title'>Content</Accordion.Section>
+      </Accordion>
+    );
+    expect([...sectionDivCount(accordion)]).toEqual([4, 4, 6]);
```
a-coruble, line 106: We could do more relevant by simulating a click on a section (ex: 1st section) that was not opened initially and check that the result is equal to `[4, 6, 4]`.

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,100 @@
...
+  clickable = true,
+  children,
+  sectionContainerStyle,
+  sectionContentClassName
+}: AccordionSectionProps) => {
+  const [isOpen, setIsOpen] = useAccordionOpen(open);
+
+  const onHeaderClick = () => {
+    if (!clickable) return;
+    setIsOpen(!isOpen);
```
a-coruble, line 39: Not sure here, as it's a custom hook based on a state one, but if we update from previous state, we should prefer using the function form ([here](https://stackoverflow.com/questions/48209452/when-to-use-functional-setstate or [here](https://medium.com/@wisecobbler/using-a-function-in-setstate-instead-of-an-object-1f5cfd6e55d1)). I don't know what that would imply though

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,100 @@
...
+  };
+
+  let Header = header ? (
+    <div onClick={onHeaderClick}>{header}</div>
+  ) : (
+    <div className={styles.AccordionSectionHeader} onClick={onHeaderClick}>
+      <div>{label}</div>
```
a-coruble, line 46: I think you could use a `header` tag here for a better semantic

sanatshah:
@sericaia @a-coruble When I began to use the accordion in leeloo, I noticed some weird behavior with it(both code and developer experience), so I pulled it in to draft to rework it! It is updated now, I kept the same PR open but let me know if you would prefer a fresh PR. 

The main code changes were removing the separate context provided to the Accordion.Sections, with instead passing the state down to them. As well as adding the behavior that when the Accordion wrapper is used, you can only specify opened sections by using the prop on the wrapper.

src/components/accordion/Section.tsx
sanatshah: Good idea, updated!

src/components/accordion/Accordion.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <Accordion.Section label='Some title' locked clickable={false}>
+        Contet
+      </Accordion.Section>
+    );
+    expect(
+      getComputedStyle(accordionSection.getDOMNode()).getPropertyValue(
+        'opacity'
+      )
+    ).toBe('0.5');
+    accordionSection.simulate('click');
```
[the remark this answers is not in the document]
sanatshah: Updated the test prior, "should render a standlone section that is locked but still clickable", to use a mock function. For this one in particular, it would not quite work since the button would never get rendered to the dom, since the section cannot be toggled

src/components/accordion/Accordion.spec.tsx
sanatshah: The test is testing more that when the accordion.sections are wrapped in an accordion, it has full control over the open states of the sections.

src/components/accordion/Section.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <div
+      className={styles.AccordionSectionContainer}
+      style={{ ...containerStyle }}
+    >
+      {Header}
+      {isOpen && (
+        <div
+          className={contentClass}
+          style={{ pointerEvents: locked ? 'none' : 'auto' }}
```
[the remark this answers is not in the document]
sanatshah: For sure, updated thanks!

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
+import * as React from 'react';
+import { Section } from './Section';
+import styles from './Accordion.module.scss';
+import { getClassNames } from '@utility/cssUtils';
+
+interface AccordionState {
+  maxOneSectionOpen?: boolean;
+  allSectionsOpened?: boolean;
+  oneSectionAlwaysOpen?: boolean;
```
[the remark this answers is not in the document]
sanatshah: No worries, thanks for the review! You might be looking at an older commit? oneSectionAlwaysOpen seems to be working appropriately : if nothing is passed to opened prop, it automatically opens the first section. 

Expected behavior for maxOneSectionOpen to override allSectionsOpened. I have updated the documentation =)

src/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,142 @@
...
+  };
+
+  try {
+    //check if there is a state being provided to us
+    sectionState = useAccordionContext();
+
+    //infer an id for the section
+    //will be the index of the component in the children array
+    sectionState.childIndex++;
+    const index = sectionState.childIndex;
```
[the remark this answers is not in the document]
sanatshah: Added some documentation for open prop.  I updated the sections to be 0 based indexed, think that it makes more sense.

src/ions/icon/base/Icon.tsx
```diff
@@ -4,13 +4,36 @@ import { dashGreen00 } from '@colors';
...
+  if (typeof iconAnimated !== 'undefined') {
+    //check if the user actually passed a iconAniamted prop
+    if (iconAnimated) {
+      transition = 'all 0.1s ease-in';
+    } else {
+      transition = 'none';
+    }
+  }
+
+  return transition;
```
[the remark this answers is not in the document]
sanatshah: Thanks, updated!

PR #56, opened 2020-04-16
Feat/checkbox

pducolin:
Glad to introduce you to the new born component, the Checkbox

Here's how it looks like

image

and here's how it interacts with mouse click and tab 
checkbox

### DIFF
--- a/src/components/checkbox/Checkbox.mdx
+++ b/src/components/checkbox/Checkbox.mdx
@@ -1,0 +1,47 @@
+---
+name: Checkbox
+menu: Components
+---
+
+import { Playground, Props } from 'docz';
+import { Checkbox } from './Checkbox';
+import { CheckboxDemo } from './CheckboxDemo';
+
+# Checkbox
+
+A checkbox component
+
+## Properties
+
+<Props of={Checkbox} />
+
+## Usage
+
+### Basic usage
+
+Examples below show the `Checkbox` in its checked and unchecked states, both in enabled and disabled modes:
+
+<Playground
+  style={{
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-around'
+  }}
+>
+  <Checkbox readOnly />
+  <Checkbox defaultChecked checked readOnly />
+  <Checkbox disabled readOnly />
+  <Checkbox disabled defaultChecked checked readOnly />
+</Playground>
+
+The example below shows an interactive `Checkbox`.
+
+<Playground
+  style={{
+    display: 'flex',
+    flexDirection: 'row',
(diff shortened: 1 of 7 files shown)


src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react';
+import { getClassNames } from '@utility/cssUtils';
+import styles from './Checkbox.module.scss';
+
+const CheckboxContainer = ({
+  className,
+  children
+}: React.InputHTMLAttributes<HTMLInputElement>) => {
```
ni0ki, line 8: As we only use className & children here, I don't think we need to inherit InputHTMLAttributes.
It makes sense only for the top-level container : Checkbox, and HiddenCheckBox as we forward the native props to the native input.

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,74 @@
...
+      {...props}
+    />
+  );
+};
+
+const StyledCheckbox = ({
+  checked = false,
+  disabled = false,
+  children
+}: React.InputHTMLAttributes<HTMLInputElement>) => {
```
ni0ki, line 34: Same remark here

pducolin:
I forgot to export it 🤦‍♀️

src/components/checkbox/Checkbox.mdx
```diff
@@ -0,0 +1,47 @@
...
+  style={{
+    display: 'flex',
+    flexDirection: 'row',
+    justifyContent: 'space-around'
+  }}
+>
+  <Checkbox readOnly />
+  <Checkbox defaultChecked checked readOnly />
+  <Checkbox disabled readOnly />
+  <Checkbox disabled defaultChecked checked readOnly />
```
akamkaro, line 34: Not sure defaultChecked is needed with readOnly

src/components/checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,83 @@
...
+      type='checkbox'
+      checked={checked}
+      {...otherProps}
+    />
+  );
+};
+
+interface StyledCheckboxProps {
+  checked: boolean | undefined;
+  disabled: boolean | undefined;
```
akamkaro, line 34: I'd prefer like this :

checked?: boolean;
disabled?: boolean;

oclero: It's not exactly the same because of struct null checks mode. (See: https://github.com/Microsoft/TypeScript/issues/13195). I also prefer the `?` notation.

PR #58, opened 2020-04-16
fix(password-input): fix missing tests on Password Input

a-coruble:


src/components/password-input/PasswordInput.spec.tsx
```diff
@@ -2,67 +2,82 @@ import * as React from 'react';
...
+    const passwordInput = mount(
+      <PasswordInput
+        showPasswordTooltipText='Reveal'
+        hidePasswordTooltipText='Hide'
+        label='Password'
+        feedbackType='warning'
+        feedbackText='Some feedback'
+      />
+    );
+    expect(passwordInput).toMatchSnapshot();
```
oclero, line 39: Some tests we could do:
- Test that a <span> with 'Some feedback' is present when feedbackType isn't undefined or null
- Test that the text for Hide/Reveal state is correctly shown when the password is visible/hidden

src/components/password-input/__snapshots__/PasswordInput.spec.tsx.snap
```diff
@@ -37,9 +41,9 @@ exports[`<Button> Global render should render default 1`] = `
                   "$$typeof": Symbol(react.forward_ref),
                   "attrs": Array [],
                   "componentStyle": ComponentStyle {
-                    "componentId": "sc-bwzfXH",
+                    "componentId": "sc-EHOje",
```
sericaia, line 44: aren't these IDs going to change every time the test run?!

oclero: I think the snapshots will be re-generated only if the code in PasswordInput changes, so not every time we run the test.

sericaia: Right, but these huge snapshots are not that useful :(

src/components/password-input/PasswordInput.spec.tsx
```diff
@@ -1,68 +1,151 @@
...
-          showPasswordTooltipText='Reveal'
-          hidePasswordTooltipText='Hide'
-          label='Password'
-          feedbackType='error'
-          feedbackText='Some feedback'
-        />
-      );
-      expect(passwordInput).toMatchSnapshot();
-    });
+describe('Global render', () => {
```
sericaia, line 6: maybe `describe(`<PasswordInput />`)` ?

src/components/password-input/PasswordInput.spec.tsx
```diff
@@ -1,68 +1,151 @@
...
+        label='Password'
+        feedbackText='Simple feedback'
+      />
+    );
+
+    expect(passwordInput.find('span')).toEqual({});
+  });
+
+  it('should call function on visibility change', () => {
+    window.alert = jest.fn();
```
sericaia, line 127: why are you using `window.alert`? 

you can just do

`const onPasswordVisibilityChanged  = jest.fn()`

then add it to the component and verify that it was called.

src/components/password-input/PasswordInput.spec.tsx
a-coruble: Yes, it would be better !

src/components/password-input/PasswordInput.spec.tsx
a-coruble: Indeed, thank you, saw that in another test in the Button component, as it's the first time I'm doing React testing with Jest. I'm going to change that

src/components/password-input/PasswordInput.spec.tsx
a-coruble: Changed !

src/components/password-input/PasswordInput.spec.tsx
a-coruble: Changed it !

src/components/password-input/__snapshots__/PasswordInput.spec.tsx.snap
pducolin: Just fell in this issue, solved a conflict in the `components/index.tsx` and ended up with new ids

pducolin: Is it possible to have snapshot test discarding style ids? (asking google while asking here)

pducolin: https://kevinjalbert.com/jest-snapshots-reducing-styled-jsx-noise/

PR #59, opened 2020-04-16
Feature/icons wrapper

sericaia:
This is a utility component that adds the icon name near the component icon so it is easier to understand which is the icon to pick

image


src/utility/icons-wrapper/IconsWrapper.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const type = (component as React.ReactElement).type as React.ComponentClass;
+  return type.name;
+};
+
+interface IconsWrapperProps {
+  children: React.ReactElement[];
+}
+// This is a utility to show components and their name side by side in the component library
+export const IconsWrapper = ({ children }: IconsWrapperProps) => {
+  return (
```
oclero, line 14: Since there is only one element, you can skip writing `return (...)`

PR #61, opened 2020-04-17
Fix/logo

sericaia:
Adds logo default values

```
* dimensions: 135x25 px
* default color: --dash-green-00
```

The reasoning to add this is that it (1) does not make much sense to add everywhere this `color`, `width`, `height` values if they are the de facto values and (2) this way we also prevent errors. 

We also guarantee that it can be changed if really necessary with new props.


src/components/logo/Logo.tsx
```diff
@@ -1,11 +1,18 @@
 import * as React from 'react';
 import { SvgWrapper } from '@ions/icon/wrapper/SvgWrapper';
 import { IconProps } from '@ions/icon/types';
+import { dashGreen00 } from '@colors';
+
+const defaultProps = {
```
maryiam, line 6: Why don't you set the default values while destructing the props in `Logo` Component?

sericaia: it seemed more complex and less readable

pducolin: I would have set default values the destructured props too, but that's personal taste

sericaia:
I noticed something else which is we have actually 3 logos: Small icon (favicon), normal icon (the one added here) and business icon.

Should this Logo component have them all with different defaults for all of them and a 
`type = small | normal | business` or should we have completely different components for them?

Either is fine for me.

pducolin:
> I noticed something else which is we have actually 3 logos: Small icon (favicon), normal icon (the one added here) and business icon.
> 
> Should this Logo component have them all with different defaults for all of them and a
> `type = small | normal | business` or should we have completely different components for them?
> 
> Either is fine for me.
> 
> @pducolin @maryiam @fmatias-dashlane @ni0ki

yes yes yes!!

PR #68, opened 2020-04-21
chore: add tests for getClassNames fonction

oclero:


pducolin:
This could be a good candidate for property based tests

PR #70, opened 2020-04-22
Feature/fluid width password input strength

a-coruble:
image
image
image
image

### DIFF
--- a/src/components/password-input/PasswordInput.module.scss
+++ b/src/components/password-input/PasswordInput.module.scss
@@ -10,6 +10,7 @@
   display: flex;
   flex-direction: column;
   align-items: flex-start;
+  width: 100%;
 
   .label {
     @include typography.body-small;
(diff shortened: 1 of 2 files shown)


a-coruble:
As I was working to implement ZXCVBN and the new Password Field / Password Strength, I realised that my widths were not flexible, and for that, I changed this behavior here on both components !

bastienGranger:
@a-coruble it is not well aligned :) 

Screen Shot 2020-04-22 at 10 44 05 AM

I have a question on that. It is cool having the messages "keep going" "That's better" It would be more helpful to have an help on "how can I can improve my password" instead

a-coruble:
@bastienGranger It is indeed helpful, but as we don't handle any business logic on this lib, we exposed the prop for the text and those tips / advices are up to the implementation (that's what I'm doing internally while I implement ZXCVBN). So this is the expected behaviour for this component :) I'll fix the alignment by tomorrow !

a-coruble:
Changed the alignment of the text to appear at the same position than the start of the gauge 
image as per @bastienGranger comment.

PR #71, opened 2020-04-23
fix(text-input): edit border colors when focused and not to match design

a-coruble:
I noticed that the focus border color and the idle border color were not the same than the ones on the Figma file, so quick update to this component.

### DIFF
--- a/src/components/text-input/TextInput.spec.tsx
+++ b/src/components/text-input/TextInput.spec.tsx
@@ -41,5 +41,16 @@ describe('<TextInput>', () => {
 
       expect(textInput).toMatchSnapshot();
     });
+
+    it('should render with fluid class', () => {
+      const textInput = mount(<TextInput fluid />);
+
+      expect(textInput).toMatchSnapshot();
+      expect(textInput.props().fluid).toBeTruthy();
+      expect(textInput.find('input').getElement().props).toHaveProperty(
+        'className',
+        'input fluid'
+      );
+    });
   });
 });
--- a/src/components/text-input/TextInput.spec.tsx
+++ b/src/components/text-input/TextInput.spec.tsx
@@ -43,13 +43,13 @@ describe('<TextInput>', () => {
     });
 
     it('should render with fluid class', () => {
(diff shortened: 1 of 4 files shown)


a-coruble:
Also, while I did the fix for the colors, I added the `fluid` prop to allow for the field to take the whole width of it's container.

src/components/text-input/TextInput.spec.tsx
```diff
@@ -41,5 +41,16 @@ describe('<TextInput>', () => {
 
       expect(textInput).toMatchSnapshot();
     });
+
+    it('should render with fluid class', () => {
+      const textInput = mount(<TextInput fluid />);
+
+      expect(textInput).toMatchSnapshot();
+      expect(textInput.props().fluid).toBeTruthy();
```
adelego, line 49: Not sure about this line, it's not providing us with any info and it can't fail. What were your intentions?

src/components/text-input/TextInput.tsx
```diff
@@ -7,23 +7,39 @@ type _ForbiddenProps = 'size' | 'prefix' | 'type';
...
 export const TextInput: React.FC<TextInputProps> = ({
   feedbackType,
   feedbackText,
+  fluid = false,
   ...htmlInputProps
 }) => {
   const feedbackCssClass =
     feedbackType && styles[feedbackType] ? styles[feedbackType] : '';
 
+  const fluidCssClass = fluid && styles.fluid ? styles.fluid : '';
```
adelego, line 34: Same comment as in your previous PR, `getClassNames` has a specific syntax for this

src/components/text-input/TextInput.spec.tsx
a-coruble: Changed this line in favour of a `toEqual(true)` as I saw that test on other PRs

src/components/text-input/TextInput.tsx
a-coruble: Changed to use the specific syntax

PR #72, opened 2020-04-23
fix: add button font weight

akamkaro:
The button font weight is now correct

### DIFF
--- a/src/components/button/Button.module.scss
+++ b/src/components/button/Button.module.scss
@@ -6,9 +6,6 @@
   border-style: solid;
   align-items: center;
   border-radius: 4px;
-  font-family: 'Public Sans';
-  font-style: normal;
-  font-weight: 600;
   justify-content: center;
   outline: none;
   transition: all 0.15s;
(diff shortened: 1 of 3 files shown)


oclero:
A question more for designers: should they add a special case for buttons in typgraphy?

src/components/button/Button.module.scss
```diff
@@ -25,19 +26,22 @@
   &.small {
     padding: 8px 16px;
     height: 32px;
-    @include typography.body-small;
```
adelego, line 28: Why did you remove this?

oclero: My guess would be because the buttton does not use the values un `typography.body-small`. Maybe we should have a `typography.button` instead of hardcoded values? (see my other comment)

pducolin:
> A question more for designers: should they add a special case for buttons in typgraphy?

better ask this on figma

src/components/button/Button.module.scss
pducolin: Indeed, button doesn't use typography body small - just a subset. I agree it should be a new typography, adding comment on figma to get designer feedback

akamkaro: The font-weight in @mixin body-small was not the good one.
And as it would be only used for the buttons, it is better in my opinion not to add styles in typography.scss, which is a file for shared styles.

oclero: I disagree. I would prefer to not have hardcoded fonts and sizes in the button's code, because if we decide to change, it'll be harder to trace if we have to go into every different component.

oclero:
> > A question more for designers: should they add a special case for buttons in typgraphy?
> 
> better ask this on figma

I've seen you've asked! thanks

src/components/button/Button.module.scss
akamkaro: Ok done :)

PR #73, opened 2020-04-23
feat(password-input): add fluid prop to take 100% of the container width

a-coruble:


src/components/password-input/PasswordInput.tsx
```diff
@@ -67,19 +73,28 @@ export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
 
   const styledClassName = (feedbackType && styles[feedbackType]) || '';
 
+  const fluidClassName = (fluid && styles.fluid) || '';
+
   const RevealOrHideIcon = showPassword ? HideIcon : RevealIcon;
 
   return (
-    <div className={styles.root}>
+    <div className={getClassNames(styles.root, fluidClassName)}>
```
adelego, line 81: You can also write:
```
getClassNames(styles.root, {[styles.fluid]: fluid})
```
Did you do it this way to avoid repeating `{[styles.fluid]: fluid}` a couple of lines below?

src/components/password-input/PasswordInput.tsx
```diff
@@ -21,6 +21,11 @@ export interface PasswordInputProps
    * The feedback nature of the input
    */
   feedbackType?: PasswordInputFeedbackType;
+  /**
+   * Make this field take 100% of the width of it's container
+   * @default false
+   */
+  fluid?: boolean;
```
adelego, line 28: I'm not sure about the naming, fluid doesn't hold a lot of meaning to me. Why not `fullWidth`?

src/components/password-input/PasswordInput.tsx
a-coruble: Yes, as I’m using it multiple times, I preferred going with this implementation

pducolin:
I agree on changing `fluid` to something more explicit

src/components/password-input/PasswordInput.tsx
oclero: It's not semantically the same. I prefer @adelego 's way, but it's not really important anyway

src/components/password-input/PasswordInput.tsx
a-coruble: I based myself on [React Semantic UI](https://react.semantic-ui.com/elements/input/#variations-fluid) which is the UI library I'm the most familiar with, but you're right, `fullWidth` is clearer, changed it for my 3 PRs

PR #79, opened 2020-04-27
Feat/colored password component

b-jan:
I only implemented snapshot tests.
Here is a screenshot of the documentation (Light / Night modes)

image
image

Here is an example of use in the Extension Popup:

image


src/components/colored-password/ColoredPassword.mdx
```diff
@@ -0,0 +1,34 @@
...
+<Props of={ColoredPassword} />
+
+## Usage
+
+### Basic usage
+
+Examples below show usage of the Colored Password
+
+<Playground
+  style={{
```
maryiam, line 24: I think that you need to pass a `preview` attribute now after the upgrade of Docz? (see [this example ](https://raw.githubusercontent.com/Dashlane/ui-components/master/src/components/button/Button.mdx)). You can also use `className` instead of style.
cc @ni0ki

src/components/colored-password/ColoredPassword.spec.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { ColoredPassword } from './ColoredPassword';
+
+describe('<ColoredPassword />', () => {
+  it('should render default', () => {
+    const passwordInput = mount(
+      <ColoredPassword value='D4$hlan3*feat$DuNk-C1ub' />
+    );
+
+    expect(passwordInput).toMatchSnapshot();
```
maryiam, line 12: Is it enough? :)

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const checkIfCharacterMatchesCategory = (
+  character: string,
+  category: StringCategory
+): boolean => {
+  return stringCategoryRegExp[category].test(character);
+};
+
+const getCharacterMatchingCategory = (character: string): StringCategory => {
+  if (checkIfCharacterMatchesCategory(character, 'letter')) {
```
maryiam, line 39: Can we use a `switch` here instead? I think it gives better readability :)

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+  id?: string;
+}
+
+const stringCategoryRegExp = {
+  letter: RegExp('[A-Za-z]'),
+  number: RegExp('[0-9]'),
+  special: RegExp('[^0-9A-Za-z]')
+};
+
+const checkIfCharacterMatchesCategory = (
```
maryiam, line 31: What do you think of moving `checkIfCharacterMatchesCategory` and `getCharacterMatchingCategory ` into a separate helper?

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+      if (
+        checkIfCharacterMatchesCategory(
+          currentPasswordCharacter,
+          currentCategory
+        )
+      ) {
+        currentCategorizedText = currentCategorizedText.concat(
+          currentPasswordCharacter
+        );
+      } else {
```
maryiam, line 71: Will you push an empty string in your `passwordSplit` if the first character doesn’t match the default category `letter`? It will not impact the UI I guess since you display an empty string but maybe we can improve this?

src/components/colored-password/ColoredPassword.mdx
ni0ki: Yep, if you want to pass a specific style / className to the preview part of the playground, please pass it like that :
```javascript
<MyComponent style={{preview: {color: 'red', [...]} className={{preview: 'myClas'}} />

I need to update the doc about this.

File: src/components/colored-password/ColoredPassword.tsx [hunk 6]
```diff
@@ -0,0 +1,107 @@
...
+  } else {
+    return 'special';
+  }
+};
+
+export const ColoredPassword: React.FC<ColoredPasswordProps> = ({
+  id,
+  value
+}) => {
+  const getPasswordSplitInStringCategories = () => {
```
Domino9697, line 52: useCallback ?

Domino9697: Maybe we could put this function in the helper as well ?

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+    passwordSplit.push({
+      text: currentCategorizedText,
+      type: currentCategory
+    });
+
+    return passwordSplit;
+  };
+
+  const coloredPasswordCategories = getPasswordSplitInStringCategories().map(
+    (values: CategorizedString, key) => (
```
Domino9697, line 95: Do we need to type the values ?

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    passwordSplit.push({
+      text: currentCategorizedText,
+      type: currentCategory
+    });
+
+    return passwordSplit;
+  };
+
+  const coloredPasswordCategories = getPasswordSplitInStringCategories().map(
```
Domino9697, line 94: I would put a useMemo here

src/components/colored-password/ColoredPassword.spec.tsx
oclero: You could write a test with a simpler password, that contains all the character types, and check if each character type is treated correctly.

src/components/colored-password/ColoredPassword.module.scss
```diff
@@ -0,0 +1,27 @@
+@use '../../colors/colors';
+@use '../../typography/typography';
+
+.password {
+  @include typography.body-medium;
+  font-family: inherit;
```
adelego, line 6: Shouldn't it be always "mono" here rather than inherit?

src/components/colored-password/ColoredPassword.mdx
b-jan: Thanks I fixed it.

src/components/colored-password/ColoredPassword.tsx
b-jan: yes good idea

src/components/colored-password/ColoredPassword.module.scss
oclero: Do we have something for monospace text in typography?

src/components/colored-password/ColoredPassword.tsx
b-jan: I moved it to helpers, and I added a types file too.

src/components/colored-password/ColoredPassword.tsx
b-jan: Please tell me if I'm wrong but I don't see how to use a switch here.
I still need to test if this the matching category is 'letter', 'number' or 'special'.

src/components/colored-password/ColoredPassword.spec.tsx
b-jan: I'll do that !

src/components/colored-password/ColoredPassword.tsx
b-jan: I am now categorizing the first character outside of the loop. It fixes that edge case.

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+interface CategorizedString {
+  text: string;
+  type: StringCategory;
+}
+
+export interface ColoredPasswordProps {
+  /**
+   * The password value to display with colors
+   */
+  value: string;
```
oclero, line 18: React noob question: why this value is passed as a Prop value and not as text like this?
```tsx
<ColoredPassword>{passwordValue}</ColoredPassword>
```

src/components/colored-password/ColoredPassword.module.scss
b-jan: image

How is working the import of font family in the library?

src/components/colored-password/ColoredPassword.tsx
b-jan: If you pass it like this, you can use it as a `children` prop (which is prefered for JSX). This is not our case because I need to apply a transformation on `passwordValue` to display different colors for the different characters. Is is clear?

src/components/colored-password/ColoredPassword.module.scss
maryiam: It supposes that the client imports the fonts on his side (for now)

src/components/colored-password/ColoredPassword.tsx
maryiam: Hm, I see. Maybe we can simplify the logic by having something like:
```js
for (const [category, regex] of Object.entries(stringCategoryRegExp)) {
    if (regex.test(char)) {
        return category;
    }
}
```

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+    return 'special';
+  }
+};
+
+export const ColoredPassword: React.FC<ColoredPasswordProps> = ({
+  id,
+  value
+}) => {
+  const getPasswordSplitInStringCategories = () => {
+    const passwordSplit: CategorizedString[] = [];
```
oclero: Another question :) Why don't you put each character in its own category?

b-jan: I'd like to group them to prevent having useless `<span>`tags.

src/components/colored-password/ColoredPassword.tsx
```diff
@@ -0,0 +1,107 @@
...
+   * The password value to display with colors
+   */
+  value: string;
+  /**
+   * The id to link with when associated to a form label
+   */
+  id?: string;
+}
+
+const stringCategoryRegExp = {
```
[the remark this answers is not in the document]
ni0ki: Will be a bit slower, but I think this form is more readable than number comparisons

src/components/colored-password/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+};
+
+export const getValueSplitInStringCategories = (
+  value: string
+): CategorizedString[] => {
+  if (value === '') {
+    return [];
+  }
+
+  const passwordSplit: CategorizedString[] = [];
```
oclero, line 33: I've played a bit to simplify this algorithm, here is a suggestion (JS, to make it more concise to read here):

```js
const getType = character => {
  const code = character.codePointAt(0)
  if (code >= 65 && code <= 90 || code >= 97 && code <= 122) {
    return 'letter' // A-Z or a-z
  } else if (code >= 48 && code <= 57) {
    return 'number' // 0-9
  } else {
    return 'special'
  }
}

const getGroups2 = password => {
  let groups = []
  let currentGroup = null

  for (let i = 0; i < password.length; i++) {
    const character = password[i]
    const type = getType(character)

    if (currentGroup && type === currentGroup.type) {
      currentGroup.text += character
    } else {
      currentGroup = {
        type: type,
        text: character,
      }
      groups.push(currentGroup)
    }
  }

  return groups
}
```

src/components/colored-password/ColoredPassword.tsx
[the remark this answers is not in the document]
b-jan: Thanks I did not know that way of doing! But here I still prefer the Regexp for readability.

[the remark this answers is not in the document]
oclero: Rexexp uses character codes underneath

src/components/colored-password/helpers.ts
b-jan: I never used that way of doing, with character codes! Thanks for letting me know!!!

oclero: And no need to have a special case for the first char, thanks to the ref of the group ;)

src/components/colored-password/ColoredPassword.tsx
b-jan: I like your solution, the problem is that in typescript, here is how I should implement it:

```
const getCharacterMatchingCategory = (character: string): StringCategory => {
  for (const [category, regExp] of Object.entries(stringCategoryRegExp) as [
    StringCategory,
    RegExp
  ][]) {
    if (regExp.test(character)) {
      return category;
    }
  }
  return 'letter';
};
```

That's why I prefer my former solution that gives better readability

oclero: Don't use an object with regexps, you're complicating the problem :) See my other comment with the algorihm improvement

src/components/colored-password/ColoredPassword.module.scss
```diff
@@ -1,9 +1,9 @@
 @use '../../colors/colors';
-@use '../../typography/typography';
 
 .password {
-  @include typography.body-medium;
-  font-family: inherit;
+  font-family: 'Space Mono';
```
oclero, line 4: Another suggestion would be to add a new item in typography, as `typography.monospace` or something like this. So we'll be able to re-use it and font families are not hardcoded in the components' css files.

src/components/colored-password/helpers.ts
```diff
@@ -1,69 +1,40 @@
...
-
-  const passwordSplit: CategorizedString[] = [];
-
-  // we start with the first character
-  let currentCategorizedText = value[0];
-  let currentCategory: StringCategory = getCharacterMatchingCategory(
-    currentCategorizedText
-  );
+  const groups: CategorizedString[] = [];
+  let currentGroup: CategorizedString = {
```
oclero, line 20: It won't work if `currentGroup` is not null at the first iteration of the loop.

src/components/colored-password/types.ts
```diff
@@ -2,5 +2,5 @@ export type StringCategory = 'letter' | 'number' | 'special';
 
 export interface CategorizedString {
   text: string;
-  type: StringCategory;
+  category?: StringCategory;
```
oclero, line 5: You can keep `category` as mandatory, but instead in the algorithm, make `currentGroup` equals `null` at start.

b-jan: it means I should change the type StringCategory with null

src/components/colored-password/ColoredPassword.module.scss
b-jan: Done

src/components/colored-password/helpers.ts
b-jan: ok as seen together ;)

PR #80, opened 2020-04-27
feat(sidenav): implement sidenav

Alexandre-Herve:
A left hand side, fixed-width, vertical, dark navigation panel.


src/components/sidenav/Sidenav.module.scss
```diff
@@ -0,0 +1,10 @@
+@use '../../colors/colors';
+
+.sidenav {
+  background-color: colors.$dash-green-00;
+  bottom: 0;
```
pducolin, line 5: you order props alphabetically? (out of curiosity)

Alexandre-Herve: Ah yes I'm used to doing that, also it's debattable for css as it splits appart stuff that belong together somethimes

src/components/sidenav/Sidenav.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Sidenav } from './Sidenav';
+
+describe('<Sidenav>', () => {
+  describe('Global render', () => {
+    it('should be styled', () => {
+      const wrapper = mount(<Sidenav></Sidenav>);
+      const hasClass = wrapper
+        .find('nav')
+        .first()
+        .hasClass('sidenav');
```
oclero, line 13: Does this test actually pass? Because the `<nav/>` element will have a unique id as a class name (something like `"sidenav-fd5ba2"`).

src/components/sidenav/Sidenav.module.scss
akamkaro: I would be a great improvement to order props alphabetically !

oclero: I don't know if that's the best. As @Alexandre-Herve said, it splits apart stuff that should be grouped together. There is a debate.

src/components/sidenav/Sidenav.spec.tsx
Alexandre-Herve: it does, I think that enzyme doesn't namespaces the classes on this test setup. The rendered class is indeed "sidenav"

pducolin:
It's weird that you don't have the docs commits. Normally when adding mdx file the live documentation website is rebuilt locally and pushed in a separate commit

maryiam:
@Alexandre-Herve Did you see errors on CLI with your commit? There is a post-commit hook:
```json
      "post-commit": "yarn exec-docz-update",
```
that should have been triggered after your commit (it checks if you committed any `.mdx` files and creates a new comment with the documentation build).
I'm not sure why it's not working here. Can you check this please or manually build documentation at worst using: 
```bash
yarn exec-docz-update
```

Alexandre-Herve:
@maryiam did not see any error, I'll run it by hand though ok!

PR #81, opened 2020-04-28
Feature/implement sidenav subcomponents

Alexandre-Herve:
implementing header, body and link subcomponents  for the sidenav.


src/components/sidenav/Sidenav.mdx
```diff
@@ -13,7 +17,16 @@ A vertical side navigation panel.
 ## Usage
 
 <Playground>
-  <Sidenav>
-    <div style={{ color: 'white' }}>Nav content</div>
-  </Sidenav>
+  <div style={{ height: 400, position: 'relative' }}>
```
oclero, line 20: You can directly customize the Playground's style

Alexandre-Herve: ok cool!

src/components/sidenav/sidenav-link/SidenavLink.module.scss
```diff
@@ -0,0 +1,26 @@
...
+  padding: 16px 24px;
+  text-decoration: none;
+
+  .sidenavLabel {
+    color: colors.$dash-green-04;
+    padding-left: 16px;
+  }
+
+  .sidenavIcon svg {
+    fill: white;
```
oclero, line 19: Maybe `colors.$white` ?

Alexandre-Herve: uh yeah I used it somewhere else but in the end I wonder if this is worth it?

oclero: I don't know ^^ it comes from designers

ni0ki: I understand it might looks dumb, but if one day, for any reason, our "white" background goes gray-ish or something like this, it will be easier to change / track down the places where this is used : )

Alexandre-Herve: Yeah but the truth is that you can never take white and change it, because white is white: you would go to the place that need to change one by one and change their color, be it 'white' or '$white', but you would not change white itself :)

oclero: I totally agree. "White" is not semantic, and something like "foreground" should be used instead. But we use the names from Figma to be aligned with the designers.

Alexandre-Herve: I changed though :)

PR #83, opened 2020-04-29
Components/thumbnail

pducolin:
Here is the Thumbnail. This implementation is pretty much a copy/paste/clean the 5 levels of divs of the one in extension project.
image

I have a RFC: instead of passing `text` or `iconSource` I would pass a children.
Pros: thumbnail could be used for other data types
Cons: user dev would need to have an `IconWrapper` somewhere - could be a separate component
```
interface ThumbnailProps {
  size: ThumbnailSize;
  backgroundColor: string;
  children: React.Node;
}

<div
      className={getClassNames(styles.container, styles[size])}
      style={{ backgroundColor: backgroundColor }}
    >
      {children}
    </div>
```
What do you think?


ni0ki:
> 
> What do you think?

No strong opinion here, it's as often, a matter of "Shall we enforce or allow ?".
I'm wondering, what could be a practical usage of a thumbnail of something other than text or image ?
If no other needs arise for now, I'd keep it this way, it's simple and I like it.

src/components/thumbnail/Thumbnail.module.scss
```diff
@@ -0,0 +1,52 @@
...
+  &.small {
+    width: 48px;
+    height: 32px;
+  }
+  &.medium {
+    width: 96px;
+    height: 64px;
+  }
+  &.large {
+    width: 147px;
```
ni0ki, line 21: 147 seems quite unusal as a value ?

pducolin:
> I'm wondering, what could be a practical usage of a thumbnail of something other than text or image ?

Secure notes, Payments and Personal Info. In the you-know-which project we used the same component for all data types

src/components/thumbnail/Thumbnail.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+      const thumbnail = mount(
+        <Thumbnail size='large' backgroundColor='gray' text='Logo' />
+      );
+
+      expect(thumbnail).toMatchSnapshot();
+    });
+  });
+
+  describe('Yellow render', () => {
+    it('should render yello', () => {
```
Domino9697, line 38: typo in yellow

pducolin: It's slang. Kidding, fixing it

src/components/thumbnail/Thumbnail.module.scss
pducolin: good catch!

pducolin: ehm no, it's actually 147 in figma

src/components/thumbnail/Thumbnail.module.scss
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.letters {
+  color: colors.$white;
+  text-transform: capitalize;
+  font-family: 'Public Sans';
+  font-weight: 600;
+  user-select: none;
+  text-align: center;
+  background-size: 46px 30px;
```
oclero, line 33: Are we sure they won't extend out of the container ? (with something like overflow or text-overflow ?) It's just in case. I don't know if it's needed.

pducolin: I used values from the extension's thumbnail. IMO we should use what I did for the icon
```
  background-size: contain;
  background-position: center;
  background-repeat: no-repeat;
```
and use margins in the child (be it text or icon)

src/components/thumbnail/Thumbnail.tsx
```diff
@@ -0,0 +1,52 @@
...
+  size: ThumbnailSize;
+  iconSource?: string;
+  text: string;
+}
+
+const Logo = ({ iconSource, text, size }: LogoProps) => {
+  return (
+    <>
+      {iconSource ? (
+        <span
```
oclero, line 41: By curiosity, why not an img tag?

src/components/thumbnail/Thumbnail.module.scss
ni0ki: Alright then, good for me.

src/components/thumbnail/Thumbnail.tsx
pducolin: I don't know, this is part of the copy/paste/don't touch it because you never know

src/components/thumbnail/Thumbnail.module.scss
pducolin: it works!

PR #85, opened 2020-04-29
refactor(password-strength): unify indicators into a single progress

a-coruble:
This is how it will look associated with a Password Input :
image

And here are the variations :
image

And this is the behaviour when the value changes : 
CleanShot 2020-04-29 at 16 04 20

### DIFF
--- a/src/components/password-strength/PasswordStrength.module.scss
+++ b/src/components/password-strength/PasswordStrength.module.scss
@@ -6,7 +6,7 @@
   flex-direction: column;
   align-items: flex-start;
   margin-top: 8px;
-  max-width: 317px;
+  max-width: 318px;
 
   &.fullWidth {
     max-width: 100%;
@@ -24,24 +24,30 @@
       max-width: 100%;
     }
 
-    .passwordStrength {
+    .passwordStrengthBackground {
+      width: 100%;
       background-color: colors.$dash-green-05;
-      border-radius: 12px;
+      border-radius: 2px;
       height: 4px;
-      margin: 0px 4px;
-      width: 20%;
+      position: relative;
 
-      &:first-child {
-        margin-left: 0px;
+      > span {
+        display: block;
+        height: 100%;
       }
+    }
 
-      &:last-child {
-        margin-right: 0px;
-      }
+    .passwordStrength {
+      display: block;
+      border-radius: 2px;
+      height: 100%;
+      transition: width 1s ease-out;
+      -moz-transition: width 1s ease-out;
+      -webkit-transition: width 1s ease-out;
 
       &.weakest,
       &.weak {
-        background-color: colors.$red-00;
+        background-color: colors.$accessible-validator-red;
       }
 
       &.acceptable {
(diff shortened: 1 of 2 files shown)


src/components/password-strength/PasswordStrength.module.scss
```diff
@@ -24,24 +24,30 @@
...
+      border-radius: 2px;
+      height: 100%;
+      transition: width 1s ease-in-out;
+      -moz-transition: width 1s ease-in-out;
+      -webkit-transition: width 1s ease-in-out;
 
       &.weakest,
       &.weak {
-        background-color: colors.$red-00;
+        background-color: colors.$accessible-validator-red;
```
sericaia, line 50: awesome that we are already using this!

pducolin:
It looks amazing!! Love it

src/components/password-strength/PasswordStrength.module.scss
```diff
@@ -6,7 +6,7 @@
   flex-direction: column;
   align-items: flex-start;
   margin-top: 8px;
-  max-width: 317px;
+  max-width: 318px;
```
goldins, line 9: This seems very magic-numbery and dependent on context. Might be okay here but not a good practice for a general purpose component library.

pducolin: would it be better to pass it as a prop, and default it to this value?

a-coruble: This value comes from the design file, but I can remove the max-width from here if you think it’s not appropriate

pducolin: I guess it should inherit width from the password element above it

ni0ki: Agreed. Given how you compute the strength bar width, depending on the top-level container width should do the job

a-coruble: The thing is that they are not linked nor nested, so we’d have to duplicate this into Password Input and the dev would have to manage both

goldins: I would let this be width: 100%, and allow the consumer of this component to set the width of the container.

a-coruble: I could do that. That would remove a prop on both components then, could be a good thing

goldins: I think it's a useful pattern to follow, especially when dealing with form layout.

oclero: Yes I agree with the width 100%. When sizes are hardcoded in Figma, it does not mean we have to hardcode them too. This component is a perfect example :)

a-coruble: Perfect, I'll change that in another PR for both Password Input and Password Strength :)

goldins: Okay just keep in mind that it would technically be a breaking change

PR #86, opened 2020-04-29
Feature/implement card component

lhoursquentin:
Very basic white background container with a thin border
card-preview
We'll probably add rows/title/... subcomponents to this in the future


src/components/card/Card.module.scss
```diff
@@ -0,0 +1,40 @@
+@use '../../colors/colors';
+@use '../../typography/typography';
+
+.card {
+  background: #ffffff;
```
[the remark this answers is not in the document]
lhoursquentin: ah you are right thanks for spotting that, will fix it

src/components/card/Card.module.scss
```diff
@@ -0,0 +1,40 @@
...
+  border-radius: 8px;
+  border-style: solid;
+  border-width: 1px;
+  box-sizing: border-box;
+  color: colors.$black;
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  margin: 16px;
+  min-width: 150px;
```
akamkaro, line 16: Ordered alphabetically <3

src/components/card/Card.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import { getClassNames } from '@utility/cssUtils';
+import styles from './Card.module.scss';
+
+type CardSize = 'small' | 'large';
```
Alexandre-Herve, line 5: Mmh I'm not a fan of this API at all. There will most likely be as many card sizes as designs.
I would rather have let the dev using the card decide the width and height with props, I don't really see why we would have a limited amount of sizes.

src/components/card/Card.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export interface CardProps extends React.HTMLAttributes<HTMLDivElement> {
+  size?: CardSize;
+}
+
+export const Card = (props: CardProps) => {
+  return (
+    <div
+      {...props}
```
Alexandre-Herve, line 14: In such a situation I would rather do something like this:
```
const { size, ...rest } = props
return <div {...rest} ...
```
Becauze `size` doesn't belong to the props for div.

Alexandre-Herve: In this case though what I would rather do is this:
```
const Card = (props: React.HTMLAttributes<HTMLDivElement>) => {
 const { className, children, ...rest } = props
 const computedClassName = [className || '', styles.card].join(' ')
 return (
   <div {...rest} className={computedClassName}>{children}</div>
 )
```
Or something in that taste

src/components/card/Card.tsx
lhoursquentin: Removed sizes, let's go very basic for now

src/components/card/Card.module.scss
```diff
@@ -0,0 +1,28 @@
...
+@use '../../typography/typography';
+
+.card {
+  background: colors.$white;
+  border-color: colors.$dash-green-05;
+  border-radius: 8px;
+  border-style: solid;
+  border-width: 1px;
+  box-sizing: border-box;
+  color: colors.$black;
```
Alexandre-Herve, line 11: I'd have left this for the user to decide, IMO the card is more a layout component

src/components/card/Card.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  margin: 16px;
+  min-height: 132px;
```
Alexandre-Herve, line 16: Seems a bit arbitrary. Why enforce this?

src/components/card/Card.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+import { getClassNames } from '@utility/cssUtils';
+import styles from './Card.module.scss';
+
+export const Card = (props: React.HTMLAttributes<HTMLDivElement>) => {
+  const { className, children, ...rest } = props;
+  const computedClassName = getClassNames(className ?? '', styles.card);
```
Alexandre-Herve, line 7: You don't need the `?? ''` anymore, I allowed `undefined` arguments in `getClassNames` lately

src/components/card/Card.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  min-width: 150px;
+  padding: 16px;
+  @include typography.body-medium;
+
+  /* Content */
+
+  .content {
```
Alexandre-Herve, line 23: I don't see this `.content` used anywhere?

lhoursquentin: Kept it from the step-by-step guide, removed it.

src/components/card/Card.module.scss
lhoursquentin: Talked offline, we might want to support dark mode later but let's keep this as is for now and use more specific styles to override it if needed.

PR #90, opened 2020-05-04
Feat/add bell icons

bradcush:
Icons related to a fix that was recently done for the sharing notification dropdown (https://gitlab.dashlane.com/webproduct/leeloo/-/merge_requests/3604) I might need to get a new icon from design as in the preview there seemed to be a glitch in the badge icon.


src/atoms/icons/BellBadge.tsx
```diff
@@ -0,0 +1,14 @@
...
+      <circle cx='13.5' cy='4.5' r='2.5' fill={functionalRed02} />
```
bradcush, line 11: I didn't see any icons that require 2 fill colors so this is hard-coded directly in the svg using a variable we already had for the color. Let me know if you think there is a better approach for something like this.

akamkaro: Do you think there will be other icons with several colors in the future ?

bradcush: Good question, I based on what I see already in the repo an I couldn't find one I'm not sure there will be outside of maybe these type of notification icons.

sericaia: We have another icon that uses 2 colors, the `BussinessLogo`. In that case we opted to include the `fill` attribute inside the `path`.

sericaia:
I believe you didn't use the generator, did you? I should have changed `icons-defs.json` and then run `yarn generate-icon-components`. There are some docs for that here https://dashlane.github.io/ui-components/documentation-icon-stepbystep/

bradcush:
@sericaia You're right, I'm preparing the release now so I'll do it tomorrow but I didn't see that section. I'll ping you when it's done.

src/atoms/icons/BellBadge.tsx
pducolin: Brad I would check with designers if this is the same component. By definition shared with them, Icon component has only one color - that makes me think we should rename it maybe `SingleColorIcon`, but that's long

oclero: A common pattern for this kind of situation is making two icons: one for the actual icon, and one for the small notification circle. And then you the the notification circle as a layer above, and it can be any color.

akamkaro: In my opinion we should improve the Icon Component with the possibilty to have more that one color.

oclero: It's much more complex to handle. Something like having props 'primaryColor', 'secondaryColor', etc. should be defined with the designers as specs, because SVG paths can be complex, and it would be hard to support it on every icon. We should know exactly what to implement, and where to stop the complexification. If it's absolutely needed right now, it'll break the API so we should talk about it. ^^

bradcush: So to sum up, what should we do here, are we ok with this for now but I also check with design. I was simply bringing an existing icon over from Leeloo as part of a bugfix.

bradcush:
@sericaia, @pducolin I've done as you said above and now generate the icons using the definitions file. Let me know if everything is good there, hopefully we can resolve the other conversation about icons with more than 1 color.

src/atoms/icons/BellBadge.tsx
akamkaro: For me it's ok as it is, we just have to find a solution in the future for icons with several colors.

pducolin: I would ask designers @bradcush

oclero:
I've fixed conflicts with `master` branch so @ni0ki can merge his refactor branch

bradcush:
@oclero Thank you, I've rebased since.

oclero:
@bradcush if everything's okay fr you, you can merge :+1:

bradcush:
@oclero Ok great, thank you.

src/atoms/icons/BellBadge.tsx
bradcush: @pducolin I've started the discussion with Clement and he is discussion with other designers but currently he is of the opinion that it's a very special case and we should treat it as such. (He asked how we prefer to have the icon as one SVG or two but to me I think one is ok honestly, thoughts?) I'll communicate what he says here when he gets back to me but I'm going to merge this PR if you are ok with me doing so...

pducolin: If this is a special case I would rather create a separate component - so not passing through the Icon process. Something like the logo. My concern is that once we'll have the CI, and we'll generate and automate the "monocolor" icons, this once will be overwritten

pducolin: You can also take care of creating a separate element in a second moment, personally I would forget about it

bradcush: Ok so I'll move the it into components but keep the icon as one element if I understand correctly, does that work?

pducolin: Yes, I would move it to `components/BellIcon` and remove it from the json. Sorry for this back and forward

bradcush: @pducolin I've made the change to move just the icon with two colors into `components/BellBadgeIcon`, the other is one color so it's remained in icons. cc: @oclero, @sericaia

src/atoms/icons/BellIcon.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+import { SvgWrapper } from '@ions/icon/wrapper/SvgWrapper';
+import { IconProps } from '@ions/icon/types';
+
```
oclero, line 4: Have you rebased on master? There should be a comment saying "This file has been generated automatically..."

bradcush: I've rebased about 10 times :) I had rebased but didn't regenerate the icons this time. Everything should be good now as the message you were referencing appears.

bradcush:
I need reviews again as everything has been reset :( One thing to note is that we shouldn't be building and committing built documentation files. We can have whatever we need to be built and run locally to view and use the live documentation but that should officially happen as part of the CI process so that we don't have 1 commit with over 100 changes and conflicts don't arise from such changes in the built files which is happening quite often. (Is this something that we're already thinking of changing soon in the future?)

pducolin:
Hi @bradcush, yes that's something we have considered - we reported it in our [FAQ](https://github.com/Dashlane/ui-components/blob/master/documentation/FAQ.md)

As soon as the repo is migrated to gitlab we'll finally have our CI

src/atoms/icons/BellIcon.tsx
pducolin: @oclero this icon is not generated. Actually I would have moved them out of icons. To be discussed, it's ok for now but we should keep this for mono-color icons only

bradcush: Only the other icon is two colors which has been moved out. This specific icon is mono-color, am I missing something?

pducolin: all good, my bad sorry!

PR #92, opened 2020-05-05
Refactor sidenav: better components organization and APIs

Alexandre-Herve:
I'm re-organizing the components so that they are:
- usable by default without configuration
- but still 100% configurable

Most of the components are simple html tags with a pre-defined class.
The `SidenavLinkItem` is a `SidenavItem` that styles the content of any anchor child: icon and text.

The objective is to be able to use it like that:
```
<SidenavLinkItem>
 <Link to={...}><Icon/>label</Link> 
</SidenavLinkItem>
```
So that the client using the lib can fully configure the `<Link>` behavior but still benefit of the styling.

Should I set this as breaking change? :thinking: It is, but no one uses that yet so I'm not sure that it is necessary.

Also a couple of notes:
- the mdx docz doesn't have the font familly loaded
- I'll probably make a MR to change the docz port to 3001 so that we can have leeloo running along side


oclero:
Good idea for the port!

src/components/sidenav/sidenav-item/SidenavItem.module.scss
```diff
@@ -0,0 +1,14 @@
+@use '../../../colors/colors';
+
+.sidenavItem {
+  border-radius: 4px;
+  box-sizing: border-box;
+  cursor: pointer;
+  height: 52px;
+  margin-bottom: 2px;
+
+
```
oclero, line 10: Do you use Prettier? Because these useless line returns are unnecessary. It's strange that they stayed after Prettier's job.

src/components/sidenav/sidenav-item/SidenavItem.tsx
```diff
@@ -0,0 +1,20 @@
...
+ * A simple <li> with a .sidenavItem class in addition to any other
+ * provided className.
+ * It styles the layout and hover effect of a standatd navigation item.
+ */
+export const SidenavItem: React.FC<SidenavItemProps> = props => {
+  const { children, className, ...otherProps } = props;
+  const customClassName = className ? ` ${className}` : '';
+  const classname = `${styles.sidenavItem}${customClassName}`;
+  return (
+    <li className={classname} {...otherProps}>
```
oclero, line 16: You can use `getClassName()` to avoid these lines:

```tsx
<li className={getClassName(className, styles.sidenavItem)} />
```

src/components/sidenav/sidenav-link-item/SidenavLinkItem.tsx
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * A SidenavItem with an .linkItem class in addition to .sidenavItem
+ * and any other provided class.
+ * It styles an svg icon and a label inside an <a> in a standard fashion.
+ */
+export const SidenavLinkItem: React.FC<SidenavLinkItemProps> = props => {
+  const { className, ...otherProps } = props;
+  const customClassName = className ? ` ${className}` : '';
+  const classname = `${styles.linkItem}${customClassName}`;
+  return <SidenavItem className={classname} {...otherProps} />;
```
oclero, line 16: Same remark here ;)

src/components/sidenav/sidenav-menu/SidenavMenu.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * A simple <ul> with a .sidenavMenu class in addition to any other
+ * provided className.
+ * It applies the layout of the menu.
+ */
+export const SidenavMenu: React.FC<SidenavMenuProps> = props => {
+  const { children, className, ...otherProps } = props;
+  const customClassName = className ? ` ${className}` : '';
+  const classname = `${styles.sidenavMenu}${customClassName}`;
```
oclero, line 14: And here also!

Alexandre-Herve:
Ok cool thanks, will do

src/components/sidenav/sidenav-item/SidenavItem.module.scss
Alexandre-Herve: It was not enabled on `.scss` files I'll fix

src/utility/cssUtils.ts
```diff
@@ -10,16 +10,18 @@ interface CssClasses {
...
   const result: string[] = [];
   classes.forEach(value => {
     if (value) {
       if (typeof value === 'string') {
         result.push(value);
       } else {
-        Object.entries(value).forEach(([key, value]) => {
-          if (value) {
-            result.push(key);
+        Object.entries(value).forEach(([className, useClassName]) => {
```
oclero, line 22: Good idea, it's more meaningful like this

Alexandre-Herve: Yes also I'm preparing to add the `no-shadow` eslint rule: in the shadows there be dragons!

Alexandre-Herve:
sorry @oclero I pushed again to prettify scss files

pducolin:
@Alexandre-Herve the font will be available as soon as we migrate this repo - fonts are hosted on our private npm, we can't expose the npm private address in dependencies as long as we are public

src/components/index.ts
```diff
@@ -15,3 +15,4 @@ export { PasswordStrength } from './password-strength/PasswordStrength';
 export { Checkbox } from './checkbox/Checkbox';
 export { Accordion } from './accordion/Accordion';
 export { Thumbnail } from './thumbnail/Thumbnail';
+export * from './sidenav';
```
pducolin, line 18: I prefer explicitly exporting components. This is because we might have subcomponents epoxrting symbols in sidenav folder, that we want to hide outside of the library. What do you think?

src/components/sidenav/Sidenav.mdx
```diff
@@ -19,12 +21,28 @@ A vertical side navigation panel.
...
-      <SidenavLink icon={<HelpIcon />} label={'label'} href={'#'} />
-    </SidenavBody>
+    <SidenavMenu>
+      <SidenavLinkItem><a href="#"><HelpIcon /> Link item with icon</a></SidenavLinkItem>
+      <SidenavLinkItem><a href="#">Link item without icon</a></SidenavLinkItem>
+      <SidenavLinkItem><a href="#"><HelpIcon /> Link item truncated text is too long</a></SidenavLinkItem>
+      <SidenavItem><span className="custom">Custom item</span></SidenavItem>
+    </SidenavMenu>
   </Sidenav>
+  <style dangerouslySetInnerHTML={{
```
pducolin, line 33: this sounds..dangerous 🧐 can you add a comment ?

src/components/index.ts
Alexandre-Herve: Then we should simply not export them in the index I guess. The index is what is publicly accessible.
My point is that if we don't scope as I did, this file could grow gigantic. It feels better when each top module adds one line.

src/components/sidenav/Sidenav.mdx
Alexandre-Herve: Haha no it's dangerous with dynamic content, here I'm just setting up a custom class to use within the demo, nothing to worry about (famous last words? :thinking: )

pducolin: what we did in other components was to create a `Demo` component, not exported, but this sounds fine (being picky)

PR #94, opened 2020-05-05
fix(thumbnail): crop content when it's too big

pducolin:
When you find that single property that fixes your UI 🎉


pducolin:
before
image
after
image

PR #89, opened 2020-05-04
fix(css): remove border-radius on icons

sericaia:
This removed the weird border that is added by default in all icons.

With border:
image

Without border 💯 :
image

PR #97, opened 2020-05-05
Feature/icon atom to component

ni0ki:
- Update the icons to get rid of styled-components.
- Move the files inside src/components


src/components/icons/Icons.ts
```diff
@@ -0,0 +1,54 @@
+/*-- EXPORT ICON COMPONENTS --*/
```
oclero, line 1: Why aren't these exports in a `src/components/icons/index.ts` file? So we would need to import with just `@components/icons`

ni0ki: More to avoid the index.ts
However, I agree this could make sense here. I'll update it

maryiam: We should not avoid using index.ts when it comes to export files. On the contrary, this is their main purpose. We should avoid using them to declare and create components on the other hand.

src/components/icons/SvgWrapper.spec.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
-import { mount } from 'enzyme';
+
 import { SvgWrapper } from './SvgWrapper';
+import { dashGreen00 } from '@colors';
+import { mount } from 'enzyme';
```
maryiam, line 5: We want to always keep import sorted in a certain way. The first section is always reserved for external lib imports (for instance react and enzyme here) and then internal imports (from files inside of the project). Can you put back this order, please?

src/components/icons/SvgWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+    '--icon-active-color': getIconColor('activeColor'),
+    '--icon-rotate': rotate ? `rotate(${rotate}deg)` : 'none'
+  };
+
+  return (
+    <svg
+      xmlns='http://www.w3.org/2000/svg'
+      viewBox={props.viewBox || '0 0 20 20'}
+      version='1.1'
+      className={`${styles.svgWrapper} ${
```
maryiam, line 81: Can you use here **getClassNames** and potentially compute the class names upstream? It's better for readably here.

src/components/alert/Alert.spec.tsx
```diff
@@ -1,7 +1,9 @@
 import * as React from 'react';
-import { mount, ReactWrapper } from 'enzyme';
-import { CheckCircleIcon, CloseIcon } from '@atoms';
+
 import { Alert, AlertProps } from '@components/alert/Alert';
+import { CheckCircleIcon, CloseIcon } from '@components';
+import { mount, ReactWrapper } from 'enzyme';
```
maryiam, line 5: Same comment for import order

plopfile.js
```diff
@@ -16,17 +16,17 @@ module.exports = plop => {
     actions: [
       {
         type: 'add',
-        path: 'src/atoms/icons/{{pascalCase name}}Icon.tsx',
+        path: 'src/components/icons/all/{{pascalCase name}}Icon.tsx',
```
maryiam, line 19: Is it useful to have the `all` folder?

plopfile.js
```diff
@@ -16,17 +16,17 @@ module.exports = plop => {
...
         templateFile: 'config/generator/icons/IconComponentTemplate.tsx.hbs',
         force: true
       },
       {
         type: 'append',
-        path: 'src/atoms/icons/index.ts',
+        path: 'src/components/icons/Icons.ts',
```
maryiam, line 25: As discussed in the other thread this should be `index.ts` instead

src/components/icons/SvgWrapper.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+import { Props } from './types';
+import { dashGreen00 } from '@colors';
+import styles from './SvgWrapper.module.scss';
+
+/**
+ * In order to support passing CSS variables with Typescriot
+ * https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors
+ */
+declare module 'csstype' {
```
maryiam, line 12: Can you move this in a separate type declaration file under `src/typings` please?

src/components/icons/SvgWrapper.spec.tsx
ni0ki: Weird, I had prettier activated, it shouldn't happen. Will fix !

src/components/icons/Icons.mdx
```diff
@@ -7,58 +7,58 @@ import { Playground, Props } from 'docz';
...
+import { ShieldFlagIcon } from './all/ShieldFlagIcon';
+import { TaxIcon } from './all/TaxIcon';
+import { ToolsIcon } from './all/ToolsIcon';
+import { UnlockIcon } from './all/UnlockIcon';
+import { VaultIcon } from './all/VaultIcon';
+import { WarningIcon } from './all/WarningIcon';
+import { WebIcon } from './all/WebIcon';
+import { BackIcon } from './all/BackIcon';
+import { DashlaneLogoIcon } from './all/DashlaneLogoIcon';
+import { ForwardIcon } from './all/ForwardIcon';
```
akamkaro, line 61: Please order alphabeticaly

src/components/icons/Icons.ts
```diff
@@ -0,0 +1,54 @@
...
+export { BackIcon } from './all/BackIcon';
+export { ForwardIcon } from './all/ForwardIcon';
+export { DashlaneLogoIcon } from './all/DashlaneLogoIcon';
```
akamkaro, line 4: please order alphabetically

src/components/icons/Icons.ts
oclero: Yes :) So we know that index.ts files don't contain anything interesting, they're just a way to export files.

src/components/icons/SvgWrapper.spec.tsx
ni0ki: This is Prettier's auto-formatting.
Is there some way to configure it ?
Otherwise I'm not sure I should disable it.

plopfile.js
ni0ki: I find this better so that when you want to work on SvgWrapper, its style or types, you don't have a list of 50+ files that pollutes your workspace on the left.

plopfile.js
ni0ki: :+1:

plopfile.js
oclero: I think you can do the opposite: put all icons in `icons/` and put SvgWrapper to `icons/utils`

src/components/icons/Icons.mdx
ni0ki: This file is generated. I'll see if I can update it so the list here is alphabetically ordered.

src/components/icons/SvgWrapper.tsx
ni0ki: :+1:

ni0ki:
I took the liberty of baking @sericaia icon's wrapper in the documentation with an idea that popped up during a slack discussion :

image

ni0ki:
Thanks to @sericaia icons wrapper in the documentation, I added the possibility of copying the icon code by click :
icon_copy

pducolin:
Amazing, thank you!

src/components/icons/utils/SvgWrapper.module.scss
```diff
@@ -0,0 +1,29 @@
+.svgWrapper {
+  display: flex;
+  width: var(--icon-width);
+  height: var(--icon-height);
+  outline: none;
+  fill: var(--icon-idle-color);
+  border-radius: 4px;
```
sericaia, line 7: @oclero told me you were removing this line and I closed https://github.com/Dashlane/ui-components/pull/89

oclero: Don't forget @ni0ki ;)

ni0ki: Thanks @sericaia, I'll do it for sure. Just waiting for the merge of Bradley's icon and I'll then rebase and remove these 4px :)

config/generator/icons/generateIcons.js
```diff
@@ -23,7 +23,7 @@ const generateIcon = async (name, path) => {
 
 const generateIconComponents = async () => {
   try {
-    for (let key of Object.keys(iconsDef)) {
+    for (let key of Object.keys(iconsDef).reverse()) {
```
akamkaro, line 26: This is cool to not begin with W :)

src/components/icons/utils/SvgWrapper.spec.tsx
```diff
@@ -42,4 +43,42 @@ describe('<SvgIcon>', () => {
...
+      const IconWrapper = mount(
+        <SvgWrapper title='Test'>
+          <React.Fragment>
+            <path d='123' />
+          </React.Fragment>
+        </SvgWrapper>
+      );
+
+      expect(IconWrapper).toMatchSnapshot();
+    });
```
lhoursquentin, line 58: Confused by this, how is `expect(IconWrapper).toMatchSnapshot();` checking `should set a default style if none is provided` ? Isn't this just checking that nothing changed between two sets of tests?

ni0ki: Hey @lhoursquentin, sorry, I saw your comment right after merging. Having a look at the snapshot, it seems that this test ensures that when providing no parameters, we have correct default color (#0E353D) & size (20px).

Seems good to me, because if we refer to the code :
```javascript
const DEFAULT_SIZE = 20; // px

[...]

const getIconColor = (propName: ColorProps = 'color'): string => {
    return isAnimated ? 'none' : props[propName] || props.color || dashGreen00;
  };
```

dashGreen00 / 20px are the default values if nothing is provided. So if one of this value changes, the test will fail.

lhoursquentin: No worries :) !
Agreed, but then I guess the description should be something like `should match the default style if none is provided`. Here if you change the default color I'm guessing the test is failing, which I would not expect based on the current description.

It just feels prone to errors, for instance if you do multiple changes, including slightly changing the color, you will see the test fail and might blindly update the snapshot (since you expect it to change due to the color change) without knowing that you broke the actual `default style if none` feature. This is of course a worst case scenario and since we are already only using snapshots in most places (and not breaking everything regularly) I think we should be fine.

PR #101, opened 2020-05-07
Add IconButton Component

Domino9697:
Today the button component cannot be a square button with only an icon inside. This is why we created an IconButton component.


src/components/button/Button.tsx
```diff
@@ -26,7 +28,8 @@ export const Button: React.FC<ButtonProps> = ({
         styles.button,
         styles[nature],
         styles[size],
-        styles[theme]
+        styles[theme],
+        className
```
maryiam, line 32: why don't we apply here `htmlButtonProps` instead?

Domino9697: yes agreed ! I did not know it was included :)

PR #104, opened 2020-05-11
fix: remove hard-coded stroke property from mobile icon

cynthiachan:
Removed stroke property from the MobileIcon component because there was a hard coded color in the icon and re-generated icons.


pducolin:
I would also move these icons somewhere else, they don't fit graphically with other icons.

oclero:
I agree. I like to store eveything related to app branding in the same folder (icons, logo, etc.), so we keep the icon library standardized.

src/atoms/icons/index.ts
```diff
@@ -1,8 +1,8 @@
 /*-- EXPORT ICON COMPONENTS --*/
 export { WindowsIcon } from './WindowsIcon';
 export { WebIcon } from './WebIcon';
-export { UserCircleIcon } from './UserCircleIcon';
 export { WarningIcon } from './WarningIcon';
+export { UserCircleIcon } from './UserCircleIcon';
 export { VaultIcon } from './VaultIcon';
```
akamkaro, line 6: order alphabetically please

oclero: Yes I second Matthieu on that: please keep the alphabetical order, so its easier to find the icon in the file.

cynthiachan: Added.

PR #105, opened 2020-05-11
WIP: Feat/success nature button

sericaia:
Would like some feedback about this idea as soon as possible ;)

In this PR I'm adding a new `nature: success` for the `accessible-validator-green` button. I don't have other designs yet for different selectors.

image


src/components/button/Button.module.scss
```diff
@@ -106,6 +106,23 @@
         color: colors.$dash-darker-green-00;
       }
     }
+
+    &.success {
+      color: colors.$white;
+      background-color: colors.$accessible-validator-green;
+      border-color: colors.$accessible-validator-green;
+
+      // &:focus,
```
pducolin, line 115: is it expected to have no hover nor focus?

sericaia: as mentioned in the PR description I'm waiting for designs ;)

oclero: I've noticed they're often associated and merged into one visual state in web UIs, unlike desktop UIs. :)

src/components/button/types.ts
```diff
@@ -1,3 +1,8 @@
-export type ButtonNature = 'primary' | 'secondary' | 'danger' | 'ghost';
+export type ButtonNature =
+  | 'primary'
```
oclero, line 2: Maybe we could begin to order alphabetically, if the list begins to grow.

Issue #37, opened 2020-04-01
Dropdown has a useless scrolbar on Firefox/Ubuntu.


oclero:
When running `yarn docz:dev` and watching the result in the browser (Firefox on Ubuntu), the Dropdown has undesired padding inside the dropdown, either it needs scrollbars or not.

![Screenshot from 2020-04-01 16-34-30](https://user-images.githubusercontent.com/1050389/78151139-8b2a0980-7438-11ea-9d12-7f2ba380a485.png)

PR #106, opened 2020-05-12
Feature/dropdown style update

lhoursquentin:
- Only display a scroll bar if a max height is specified (includes tests), was previously defaulting to 100px if not specified
- Fix playground display where the code block was displayed on top of the drop-down items
- Small update to the current colors and height of the drop-down items.


src/components/dropdown/DropdownElement.tsx
```diff
@@ -17,15 +17,16 @@ const Wrapper = styled.button`
   outline: none;
   display: flex;
   box-sizing: border-box;
-  height: 36px;
+  height: 40px;
```
Alexandre-Herve, line 20: This is a breaking change, you need to tag your commit as such! Do you know who is already using the dropdown?

Alexandre-Herve: Also in a way I find it weird that we are using a custom styled button when we have a button component :thinking:

src/components/dropdown/DropdownMenu.spec.tsx
```diff
@@ -1,24 +1,56 @@
 import * as React from 'react';
 import { mount } from 'enzyme';
+import { DropdownCard } from './subcomponents/DropdownCard';
 import { DropdownMenu } from './DropdownMenu';
 import { DropdownElement } from './DropdownElement';
 
-describe('<Button>', () => {
+interface DropdownMenuProps {
+  isOpen: boolean;
+  maxHeight?: number;
```
Alexandre-Herve, line 9: you can add a jsdoc to explain that defining maxHeight will lead to a scrollbar

Alexandre-Herve: Uh I did not see that it was a test file. I wouldn't expect that you define an interface for props in a test file. Weird.

src/components/dropdown/subcomponents/DropdownCard.tsx
```diff
@@ -1,4 +1,4 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
```
Alexandre-Herve, line 1: You should add the jsdoc on the maxHeight prop in this file, and import those Props in your test file rather than re-defining them

oclero:
Does it fixes https://github.com/Dashlane/ui-components/issues/37 ?

lhoursquentin:
> Does it fixes #37 ?

- The scrollbar part yes (if no max height is specified)
- The wrong placement no, but I'm not seeing this issue today.

lhoursquentin:
> * The wrong placement no, but I'm not seeing this issue today.

Correction, I do indeed see this placement issue for the very last drop-down on the left in Firefox (as specified in the issue). And I can confirm that this PR won't fix it.

oclero:
OK. So here is my idea: I edit the Github issue to reference only the scrollbar bug, so you can mark it as fixed. And I open another Github issue for the placement bug.

lhoursquentin:
> OK. So here is my idea: I edit the Github issue to reference only the scrollbar bug, so you can mark it as fixed. And I open another Github issue for the placement bug.

Sounds good to me :), thanks!

oclero:
Done!

src/components/dropdown/DropdownElement.tsx
oclero: Indeed it's weird and needs refactor

lhoursquentin: - Updated the commit message with breaking change.
- Regarding who is using it, yes I made sure we are ok with this update for the places already using it
- Agreed for the button, would you guys prefer to do the transition in this PR? Or later?

Alexandre-Herve: that's your take depending on what you want to do :shrug:

zeosamaster: fyi, PTF (Protect the Family) is also using the Dropdown component

lhoursquentin: Sounds good, to keep it simple let's do that in another PR

src/components/dropdown/subcomponents/DropdownCard.tsx
```diff
@@ -14,6 +14,10 @@ export const DropdownCard = styled.div<Props>`
   box-shadow: 0px 4px 16px rgba(0, 0, 0, 0.2);
   border-radius: 4px;
   padding: 8px 0;
-  max-height: ${props => props.maxHeight || 100}px;
-  overflow-y: scroll;
+  ${props =>
+    props.maxHeight &&
+    css`
+      max-height: ${props.maxHeight}px;
```
ni0ki, line 20: Could we have a situation where props.maxHeight is undefined ?
What are the consequences ? It seems to be an optional prop.

lhoursquentin: Yes, `props.maxHeight` could be undefined, the intent is that if `maxHeight` is not provided, then no scroll bar is added (and of course not default `max-height` property is added either, whereas we had a default `100px` max height in that case before)

ni0ki: Understood, thanks for the explanation :+1:

src/components/dropdown/DropdownMenu.spec.tsx
```diff
@@ -1,24 +1,58 @@
...
+    >;
+    return mount(
+      <div>
+        <div ref={controllerRef}>Dropdown Menu controller</div>
+        <DropdownMenu
+          controllerRef={controllerRef}
+          isOpen={props.isOpen}
+          menuMaxHeight={props.menuMaxHeight}
+        >
+          <DropdownElement onClick={() => {}}>Element 1</DropdownElement>
```
oclero, line 20: `onClick` prop could be optional (like for Button or other components).

lhoursquentin: Agreed! I'll do that as part of getting of styled-components for this dropdown

src/components/dropdown/DropdownMenu.mdx
```diff
@@ -62,6 +62,49 @@ Click Me
 
 </Playground>
 
+
+Dropdown can be defined with a max-height, in which case it will have a scroll
+bar.
+
+<Playground
+  style={{
```
Alexandre-Herve, line 70: I don't think that this works, or does it?
I think that you have to wrap your style like so for instance:
```
 <Playground style={{ preview: {
    height: '500px',
    position: 'relative',
    display: 'grid',
    gridTemplateColumns: '256px 96px',
    gridTemplateRows: '89px 1fr',
    gridGap: '0 24px',
  }}}>

```

lhoursquentin: Yes you are right, good catch, I assumed it was correct when I copied it over.
Fixed this new one and all the existing others for the dropdown.

PR #107, opened 2020-05-12
fix(icon-button): fix background and border colors

Domino9697:
Fixes the icon buttons' background color and border color to respect the figma designs.

### DIFF
--- a/src/components/icon-button/IconButton.module.scss
+++ b/src/components/icon-button/IconButton.module.scss
@@ -5,6 +5,8 @@
   min-width: unset;
   min-height: unset;
 
+  /* Size */
+
   &.small {
     height: 32px;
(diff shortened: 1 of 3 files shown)


src/components/icon-button/IconButton.module.scss
```diff
@@ -22,4 +24,31 @@
...
+
+  &.light {
+    &:focus,
+    &:hover:enabled {
+      border-color: colors.$dash-green-05;
+    }
+
+    &:active:enabled {
+      border-color: colors.$dash-green-05;
+      background-color: colors.$dash-green-05;
```
akamkaro, line 38: i'd prefer ordered alphabetically but it's a detail :)

PR #110, opened 2020-05-19
fix(types): fix type declaration import alias

Domino9697:
Fixes an issue in the ts declaration files. 

Basically the rollup configuration we have does not resolve the import alias of typescript declaration files. In other words if a dev imports only a type from a file like:

```ts
import { Placement } from '@components/Tooltip/types'
```

In the build files, we would have a .d.ts file with the same import:

 ```ts
import { Placement } from '@components/Tooltip/types'
```

This led to the ui components types to be completely broken.

This PR addresses this issue by resolving the imports.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,3 +1,4 @@
+import ttypescript from 'ttypescript';
 import postcss from 'rollup-plugin-postcss';
 import compiler from '@ampproject/rollup-plugin-closure-compiler';
 import typescript from 'rollup-plugin-typescript2';
@@ -19,7 +20,7 @@ const config = {
       plugins: [autoprefixer()]
     }),
     typescript({
-      typescript: require('typescript'),
+      typescript: ttypescript,
       objectHashIgnoreUnknownHack: true
     }),
     json({ preferConst: true })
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: rollup.config.js [hunk 1]
```diff
@@ -19,7 +20,7 @@ const config = {
       plugins: [autoprefixer()]
     }),
     typescript({
```
maryiam, line 22: Have you tried configuring `typescript` this way rather than using a plugin ?
```js
typescript({ 
      typescript: require('ttypescript'),
      tsconfigDefaults: {
        compilerOptions: {
          plugins: [
            { "transform": "typescript-transform-paths" },
            { "transform": "typescript-transform-paths", "afterDeclarations": true }
          ]
        }
      }
})
```

Domino9697: It works as well I tested it. You prefer typescript-transform-paths ?

maryiam: I think if we avoid using extra plugins, it would be better. What do you think?

Domino9697: Yes I agree ! Let me change that

PR #112, opened 2020-05-20
Fix/linting issues

sericaia:
React hooks useEffect dependencies is not being properly handled.

This commits adds linting for react hooks and fixes the current useEffect's problems.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,7 +3,14 @@
   "parserOptions": {
     "project": "./tsconfig.json"
   },
-  "plugins": ["@typescript-eslint", "prettier", "react", "json", "import"],
+  "plugins": [
+    "@typescript-eslint",
+    "prettier",
+    "react",
+    "react-hooks",
+    "json",
+    "import"
+  ],
   "extends": [
     "eslint:recommended",
     "plugin:react/recommended",
@@ -21,6 +28,9 @@
   "rules": {
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-var-requires": "error",
+    "react/no-string-refs": "warn",
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
     "prettier/prettier": "error",
     "import/no-relative-parent-imports": "error",
     "no-console": ["error", { "allow": ["error"] }],
--- a/src/components/accordion/Accordion.tsx
+++ b/src/components/accordion/Accordion.tsx
@@ -39,7 +39,7 @@ const Accordion = ({
     } else {
       setOpenedSections(new Set<number>(opened));
     }
-  }, [opened]);
+  }, [opened, allSectionsOpened, children, oneSectionAlwaysOpen]);
 
   const toggleSection = (index: number, requestedOpen: boolean) => {
     if (requestedOpen) {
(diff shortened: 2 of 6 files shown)


.eslintrc.json
```diff
@@ -21,6 +28,9 @@
   "rules": {
     "@typescript-eslint/explicit-function-return-type": "off",
     "@typescript-eslint/no-var-requires": "error",
+    "react/no-string-refs": "warn",
```
goldins, line 31: 👍  good addition too

src/utility/icons-wrapper/IconsWrapper.tsx
```diff
@@ -25,39 +25,45 @@ const successTooltip: Props = {
   placement: 'right'
 };
 
+interface IconWrapperProps {
+  child: React.ReactElement;
+}
+
+const IconWrapper = ({ child }: IconWrapperProps) => {
```
goldins, line 32: IMO wrapping `IconWrapper` in `memo` would be cleaner than using two `useCallback`s. Performance gains might be negligible in either case though. (out of scope; up to you if you want to make changes)

src/components/accordion/Accordion.tsx
```diff
@@ -39,7 +39,7 @@ const Accordion = ({
     } else {
       setOpenedSections(new Set<number>(opened));
     }
-  }, [opened]);
+  }, [opened, allSectionsOpened, children, oneSectionAlwaysOpen]);
```
sanatshah, line 42: Thanks for this!

src/components/dropdown/DropdownMenu.tsx
```diff
@@ -75,22 +84,19 @@ export const DropdownMenu: React.FC<Props> = props => {
       console.error('Error while computing DropdownMenu position', e);
       setPlacement(DEFAULT_PLACEMENT);
     }
-  });
+  }, [containerRef, controllerRef, dropdownPlacement, isOpen]);
```
sericaia, line 87: I'm not 100% sure if we shouldn't add `menuRef` here too (linter does not complain). Any thoughts @adelego ?

src/components/dropdown/DropdownMenu.tsx
```diff
@@ -38,34 +39,42 @@ const MenuWrapper = styled.div<MenuWrapperProps>`
 const DEFAULT_PLACEMENT = 'bottom';
 const POSSIBLE_PLACEMENTS: Placement[] = ['top', 'bottom'];
 
-export const DropdownMenu: React.FC<Props> = props => {
+export const DropdownMenu = ({
```
adelego, line 42: I think it's better to explicitly type components. This way if we make a mistake with the return type, typescript will point it out.
Also, you won't need to explicit the `children` props

src/components/dropdown/DropdownMenu.tsx
adelego: Yes it would be better I think

src/utility/icons-wrapper/IconsWrapper.tsx
ni0ki: On this, I'm still learning when to use one, the other, or not use them at all, thanks for the advice :+1:
Might have wrote this a bit too fast.

src/components/dropdown/DropdownMenu.tsx
sericaia: I'm typing them we way we do it in another projects. I believe I can use `React.PropsWithChildren` instead.

PR #113, opened 2020-05-26
Refactor/sidenav links

Alexandre-Herve:
- SidenavItem is now just a layout component, and SidenavLink handles all the business of styling the inner of a link.
- Add selected state and notifications
- Add collapsed state


src/components/sidenav/sidenav-link/SidenavLink.tsx
```diff
@@ -0,0 +1,52 @@
...
+  /**
+   * Enables thinner collapsed mode
+   */
+  collapsed?: boolean;
+  /**
+   * Displays a red dot indicating a pending notification on the target page
+   */
+  hasNotification?: boolean;
+  /**
+   * The icon component do display in the link
```
ni0ki, line 20: Let's be nitpicky: small typo here, "to display" instead of "do display" ?

PR #114, opened 2020-05-28
Feat/add dashlane d logo

Alexandre-Herve:
Adding this guy as a component:

snapshot571


src/components/d-logo/logo.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="48" height="64" viewBox="0 0 48 64" fill="none" xmlns="http://www.w3.org/2000/svg">
```
ni0ki, line 1: Is there a particular use case for this SVG file to be here ?

Alexandre-Herve: I'm copying the way the other logo are done (logo and business logo). I'm not sure that they should go with the icons, they are a bit different in nature

ni0ki: Sorry, my question was not clear enough. I'm OK with the "D" Logo to be a component by itself, but as we have the whole code in DLogo.tsx, why do we need this logo.svg file ?
I don't see something similar for the business logo

PR #115, opened 2020-05-28
Feat/text input inline feedback

sanatshah:
**Goal**  :  The ability to have specified right justified inline text within the input box. The text should be able to be dynamic and be of varying size and not cover the input box. And of course still support the other props.

inline

**Proposed solution** : Getting the input to size correctly for dynamic text was a bit tricky. I concluded with a solution that renders the input + another span wrapped in another div. The span is styled similar to the input and can then dynamically size up in the container. I'm not sure if this is the best approach, let me know if you have a better idea =). 

**Changes** : 
- Main change was adding a **InlineTextWrapper** function in **TextInput.tsx** that wraps the input and the added span into a div that is a row flexbox. This was needed to be able to support the outer div container functionality of the feedbackText (which was a column flexbox)
- Added css for the span, which is connected to the input :hover, .warning, .error etc. using ~, to get functionality such as this 

warningerror


oclero:
The inline feedback text does not change color when the feedback changes? It just look strange. Is it by design, from Figma?

src/components/text-input/TextInput.tsx
```diff
@@ -15,35 +16,71 @@ interface TextInputProps
    * Some text to display under the input when triggering a feedback type
    */
   feedbackText?: string;
+  /**
+   * The nature of the inline text
+   */
+  inlineTextType?: InlineTextType;
```
oclero, line 22: As said in my comment on the issue, I think it's not useful, and even error-prone, to two different "feedback types" in the same component.

src/components/text-input/TextInput.tsx
```diff
@@ -15,35 +16,71 @@ interface TextInputProps
...
+}> = ({ inlineText, inlineCss, children }) =>
+  inlineText ? (
+    <div className={styles.inlineInputContainer}>
+      {children}
+      <span className={getClassNames(styles.inlineText, inlineCss)}>
+        {inlineText}
+      </span>
+    </div>
+  ) : (
+    <>{children}</>
```
oclero, line 49: Why do you need inlineText, inlineCss and children at the same time? What's meant to be actually displayed in this component? Is it just text?

sanatshah: Not sure if I understand your question. But essentially if an inlineText was provided this function wraps the children (in this case the input component) within a row flex container, in order to share the space with our inlineText and its styling (inlineCss). If no inlineText is provided, it just returns the children

src/components/text-input/TextInput.tsx
sanatshah: Hmm yea I see what you are saying. I was intentionally giving control to the dev to determine the state of the inline text. I'll update it to follow the single feedback type

sanatshah: Updated =)

src/components/text-input/TextInput.tsx
oclero: OK I see

src/components/text-input/TextInput.module.scss
```diff
@@ -76,6 +84,59 @@
...
+  /* config */
+  white-space: nowrap;
+  text-align: right;
+
+  /* border */
+  border: 1px solid colors.$dash-green-05;
+  border-top-right-radius: 4px;
+  border-bottom-right-radius: 4px;
+  box-sizing: border-box;
+  border-left: 0px;
```
cynthiachan, line 125: We can omit px when using 0px values.

src/components/text-input/TextInput.module.scss
```diff
@@ -76,6 +84,59 @@
...
 
+.inlineInputContainer {
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+  height: 100%;
+}
+
+.inlineInput {
+  border-right: 0px;
```
cynthiachan, line 95: same here

src/components/text-input/TextInput.module.scss
sanatshah: Oh right, thanks updated!

src/components/text-input/TextInput.module.scss
sanatshah: Updated =)

PR #117, opened 2020-06-05
fix: apply html props to SvgWrapper

pducolin:
Apply html props to SvgWrapper


src/components/icons/utils/SvgWrapper.tsx
```diff
@@ -66,16 +72,17 @@ export const SvgWrapper = ({
 
   return (
     <svg
+      {...otherProps}
```
Domino9697, line 75: Awesome !

We should do this by default in other components !

PR #118, opened 2020-06-05
Feat/extension 4673 add label to checkbox

pducolin:
I refactor the checkbox while adding the label - having subcomponent resulted in lower readability of css classes


pducolin:
> While the implementation is correct, I'm not sure this is the way to go.
> We're giving the checkbox too much logic to handle IMHO, and here are the problems that pop into my mind :
> 
>     * If I want a "fixed" label, I should not have to write
> 
> 
> ```
> <CheckBox checked="MyLabel" unchecked="MyLabel" />
> ```
> 
>     * Checkbox may sometimes have a 3rd state (https://www.w3.org/TR/wai-aria-practices/examples/checkbox/checkbox-2/checkbox-2.html)
> 
> 
> I think it should be up to the parent to set the properties given the state of the checkbox.
> 
> I'd rather see something like this :
> 
> ```
> <CheckBox label="MyLabel" showLabel={true}/>
> ```
> 
> Will be a bit more tedious to manage for the parent, but I think it's more clear / robust

Might even used only `label` and show it if not empty, what do you think?

ni0ki:
Sure :)

src/components/checkbox/Checkbox.tsx
```diff
@@ -67,17 +40,30 @@ const Checkmark = () => {
   );
 };
 
+export interface CheckboxProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
+  label?: string;
```
pducolin, line 45: Done @ni0ki 🎉

src/components/checkbox/Checkbox.tsx
```diff
@@ -67,17 +40,30 @@ const Checkmark = () => {
...
-  ...props
-}: React.InputHTMLAttributes<HTMLInputElement>) => (
-  <CheckboxContainer className={className}>
-    <HiddenCheckbox checked={checked} {...props} />
-    <StyledCheckbox checked={checked} disabled={disabled}>
-      <Checkmark />
-    </StyledCheckbox>
-  </CheckboxContainer>
-);
+  label = '',
```
ni0ki, line 52: This could be left uninitialized I think.
In the render part, simply checking with
```
{label ? [...] }
```
will also be faster as we don't have to initialize an empty string

src/components/checkbox/Checkbox.tsx
ni0ki: Thanks. I think it's better this way, but I'd like to have some other opinion on this though :)

src/components/checkbox/Checkbox.tsx
```diff
@@ -67,17 +40,32 @@ const Checkmark = () => {
...
+  showLabel = false,
+  ...otherProps
+}: CheckboxProps) => {
+  return (
+    <label className={getClassNames(styles.checkboxContainer, className || '')}>
+      <input
```
Domino9697, line 57: Why is the input within the label ?

src/components/checkbox/Checkbox.tsx
```diff
@@ -67,17 +40,32 @@ const Checkmark = () => {
   );
 };
 
+export interface CheckboxProps
+  extends React.InputHTMLAttributes<HTMLInputElement> {
+  label?: string;
+  showLabel?: boolean;
```
Domino9697: Why do we have a showLabel boolean ?
Sorry maybe I just missed something ?

src/components/checkbox/Checkbox.tsx
ni0ki: It's a (valid per the W3 specs) way to associate them.
It's the same than using `<label for="[...]" />`

src/components/checkbox/Checkbox.tsx
pducolin: Removing it

src/components/checkbox/Checkbox.tsx
pducolin: No worries! Changing it

src/components/checkbox/CheckboxDemo.tsx
```diff
@@ -3,11 +3,14 @@ import { Checkbox } from './Checkbox';
...
 export const CheckboxDemo = () => {
   const [isChecked, setIsChecked] = React.useState(false);
+  const label = isChecked ? 'Checked' : 'Unchecked';
   return (
     <div>
       <Checkbox
         checked={isChecked}
         onChange={e => setIsChecked(e.currentTarget.checked)}
+        label={label}
+        showLabel={true}
```
ni0ki, line 13: This is not use anymore I think

pducolin: ouch forgot it, thank you for catching

PR #119, opened 2020-06-05
Feat/add bankaccount thumbnail

oclero:
1. I've added the thumbnail for bank accounts. It's not an icon, so I've not added it to icons, but rather to its own folder.
2. I've fixed incorrect paths in the docs for icons
3. I've sorted alphabetically the exports in the gobal index.ts

Screenshot from 2020-06-05 16-42-04


src/components/bankAccountThumbnail/BankAccountThumbnail.tsx
```diff
@@ -0,0 +1,64 @@
...
+        d='M101.743 21.4287L101.862 21.5H102H240C243.038 21.5 245.5 23.9624 245.5 27V138C245.5 141.038 243.038 143.5 240 143.5H5.99999C2.96243 143.5 0.5 141.038 0.5 138V6C0.5 2.96244 2.96243 0.5 6 0.5H66.8615L101.743 21.4287Z'
+        fill='#F5F7F7'
+        stroke='#CFD7D8'
+      />
+    </React.Fragment>
+  </svg>
+);
+
+export const BankAccountThumbnail: React.FC<BankAccountThumbnailProps> = (
+  props: BankAccountThumbnailProps
```
sericaia, line 54: you can destructure props and use `{size} : BankAccountThumbnailProps`  instead. Non blocking  ;)

src/components/bankAccountThumbnail/BankAccountThumbnail.tsx
```diff
@@ -0,0 +1,64 @@
...
+) => {
+  switch (props.size) {
+    case 'small':
+      return <BankAccountThumbnailSmall />;
+    case 'medium':
+      return <BankAccountThumbnailMedium />;
+    default:
+      return null;
```
sericaia, line 62: This can never happen right?

oclero: It can happen in the playground, in the live docs.

src/components/bankAccountThumbnail/BankAccountThumbnail.tsx
oclero: Strangely, eslint does not want me to do so. I'll let it as it is for the moment.

src/components/bankAccountThumbnail/BankAccountThumbnail.tsx
```diff
@@ -0,0 +1,64 @@
...
+    xmlns='http://www.w3.org/2000/svg'
+    version='1.1'
+    className={styles.bankAccountThumbnail}
+    viewBox='0 0 48 32'
+    style={{ width: '48px', height: '32px' }}
+  >
+    <React.Fragment>
+      <path
+        d='M16.5 3H39.9959C41.9959 3 42.9959 4 42.9959 6H19.9959L16.5 3Z'
+        fill='#CFD7D8'
```
pducolin, line 20: could we use tokens for colors?

oclero: I don't really know if we can use tokens for this, since it's illustrations. Fortunately, this one seems to use already existing colors from tokens, so I can replace them by the tokens. However, for more complex illustrations, that contain colors that are not in the brand tokens, we should let them. I'm adding that in a next commit.

pducolin: Did you check with your designer if they plan to have different colors based on bank ? We do it for the credit card for example. In that case the two colors (or are they more) could become a prop

oclero: At the moment, no color. We'll add the prop the day we might need it.

src/components/bankAccountThumbnail/BankAccountThumbnail.tsx
adelego: We can't make sure `ui-comp` users use TS and verify that `size` is always 'small' or 'medium' so we should definitely always to a default case

PR #121, opened 2020-06-08
Feat/add magic wand icon

perrywortman:
Creating a new icon component for the design related to try autofill step of the onboarding experience.


perrywortman:
@ni0ki Good question, and looks like there is conversation happening already internally on Slack around this topic. I think that our icons need some updating from design side as there are many inconsistencies around the viewboxes. Does "0 0 20 20" align with what we want moving forward?

ni0ki:
Unfortunately, I don't have the answer either. We'll have to wait for the design team feedback on this.
My only worry is that someone who wishes to use an icon shouldn't have to care about its default viewbox. This is a limitation of how we generate icon for now.

Maybe we could improve the system this way to handle multiple cases :

```javascript
"magic wand": {
    path: [...],
    viewBox: "0 0 32 32
}
```

which would translate into
```javascript
  <SvgWrapper {...props} viewBox='0 0 32'>
    <React.Fragment>
     [...] // SVG path
```

But this is far outside the scope of this PR, that's why I'm totally fine with it, as long as we pay attention for now :)

src/components/icons/Icons.mdx
```diff
@@ -130,14 +130,14 @@ playground of the icons and add some props.
...
+    <MagicWandIcon viewBox='0 0 40 40' size={30} />
     <MailIcon size={30} />
-    <MobileIcon size={30} />
+    <MobileIcon viewBox='0 0 40 40' size={30} />
     <NoNetworkIcon size={30} />
     <OpenWebsiteIcon viewBox='0 0 24 24' size={30} />
     <PasswordChangerIcon size={30} />
     <PasswordHistoryIcon size={30} />
-    <PasswordsIcon size={30} />
+    <PasswordsIcon viewBox='0 0 40 40' size={30} />
```
ni0ki, line 140: `PasswordsIcon` is used in Leeloo. (`src/webapp/onboarding/index.ts#24`) with a different viewbox.
I'm worried this changed could break the UI :|
ping @cynthiachan who worked on it

perrywortman: She aware as we're on the same team. I'm submitting an MR with all the proper viewbox sizes in leeloo as well.

ni0ki: :+1: Thanks !

PR #124, opened 2020-06-16
feat: queerify password strength

pducolin:
Pride Easter Egg 🏳️‍🌈 
image

### DIFF
--- a/src/components/password-strength/PasswordStrength.mdx
+++ b/src/components/password-strength/PasswordStrength.mdx
@@ -27,7 +27,7 @@ Examples below show usage of the Password Strength
     display: 'flex',
     flexDirection: 'column',
     justifyContent: 'space-between',
-    height: 400,
+    height: 430,
     flex: 1
   }}
 >
@@ -51,5 +51,11 @@ Examples below show usage of the Password Strength
     score={4}
     showAdditionalText={true}
     additionalText='Hurray !'
+  />
+  <PasswordStrength
+    score={4}
+    showAdditionalText={true}
+    showQueerColors={true}
+    additionalText='Colorful and diverse characters make us stronger.'
   />
 </Playground>
--- a/src/components/password-strength/PasswordStrength.module.scss
+++ b/src/components/password-strength/PasswordStrength.module.scss
@@ -49,6 +49,31 @@
       &.strong {
         background-color: colors.$accessible-validator-green;
       }
+
+      &.queer {
+        // Philadelphia Pride Flag colors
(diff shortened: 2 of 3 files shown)


src/components/password-strength/PasswordStrength.tsx
```diff
@@ -24,6 +24,10 @@ export interface PasswordStrengthProps {
    * Background of the strength indicator
    */
   backgroundColor?: string;
+  /**
+   * Use Philly Pride flag colors for score 4 strength
+   */
+  isQueerified?: false;
```
ni0ki, line 30: As we use an "active" action for the "showAdditionalText", it would be great to have an active form here as well.
`showQueerColors` or something like that ?

src/components/password-strength/PasswordStrength.mdx
```diff
@@ -52,4 +52,10 @@ Examples below show usage of the Password Strength
     showAdditionalText={true}
     additionalText='Hurray !'
   />
+  <PasswordStrength
+    score={4}
+    showAdditionalText={true}
+    isQueerified={true}
+    additionalText='We like our passwords like rainbows⏤beautiful, bold, and diverse.'
```
adelego, line 59: 😍 Love this
Should this be translated?

pducolin: It should, but it will be handled in projects that use it

src/components/password-strength/PasswordStrength.module.scss
```diff
@@ -49,6 +49,28 @@
       &.strong {
         background-color: colors.$accessible-validator-green;
       }
+
+      &.queer {
+        background: linear-gradient(
+          90deg,
+          #000000 0%,
```
sericaia, line 56: can we use dashlane's `$black` instead? is there any other colors we can use from our palette?

pducolin: I think it's risky to bind it to Dashlane's color, as color comes from https://en.wikipedia.org/wiki/Rainbow_flag_(LGBT)#/media/File:Philadelphia_Pride_Flag.svg
So, if Dashlane's black changes, we might mess up these colors. I can define some variable, or even define the gradient as a css variable for better readability, something like
--philly-pride-gradient. What do you think?

sericaia: maybe adding a comment with that URL is enough

oclero: aaah yes the colors should stay the standardized colors of the flag

PR #125, opened 2020-06-16
feat: update secondary button colors, according to latest design

pducolin:
Less funny, still useful PR to update secondary button design 
Before:
before
After:
after


oclero:
By the way, there is no generated docs?

a-coruble:
There should be a generated change for the tests I think no ? By the way, do you have screen captures of before / after ?

pducolin:
That's weird indeed

pducolin:
> There should be a generated change for the tests I think no ? By the way, do you have screen captures of before / after ?

No changes in snapshots. A snapshot doesn't include colors, for the secondary button the snapshot is ```

<Component
  nature="secondary"
>
  <button
    className="button secondary medium light"
  />
</Component>
```

Added gif of before / after

PR #128, opened 2020-06-22
add otp field

Domino9697:
A gif explains things better:

Peek 2020-06-22 10-57


src/components/otp-field/OTPClock.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    // Set the timeLeft to the real time left to catch up the possible delay
+    // Important to have a 'catch up' animation
+    if (!timeRef.current) {
+      setTimeLeft(validityEndDate - Date.now());
+    }
+
+    // Update the timeLeft each second (in sync with the 1s animation of the SVG)
+    timeRef.current = setTimeout(() => {
```
pducolin, line 54: did you consider using `setInterval` ?

src/components/otp-field/OTPClock.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+export interface OTPClockProps {
+  /**
+   * Total time the code should be valid
+   */
+  validityPeriod: number;
+  /**
+   * Unix time when the code is no longer valid
+   */
+  validityEndDate: number;
```
pducolin, line 17: Is `validityEndDate` mandatory? I'm thinking about a case where endDate is `DateTime.now() + validityPeriod`

src/components/otp-field/OTPClock.tsx
```diff
@@ -0,0 +1,90 @@
...
+  React.useEffect(() => {
+    if (timeRef.current && timeLeft < 0) {
+      timeRef.current && clearTimeout(timeRef.current);
+      onTimerEnd && onTimerEnd();
+      return;
+    }
+
+    // Set the timeLeft to the real time left to catch up the possible delay
+    // Important to have a 'catch up' animation
+    if (!timeRef.current) {
```
pducolin, line 49: I would move this at the top of the function. Don't you need to check if `timeRef` is null ?

src/components/otp-field/OTPClock.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from 'react';
+
+import { getClassNames } from '@utility/cssUtils';
+import styles from './OTPClock.module.scss';
+
+const PERIMITER = 38;
+
+export interface OTPClockProps {
```
pducolin, line 9: I would remove the otp reference

src/components/otp-field/helpers.ts
```diff
@@ -0,0 +1,3 @@
+export function formatCode(code: string): string {
```
pducolin, line 1: I watched a tuto recently about clean code, and it was warning on using `helper` files. So I'm questioning now, could this be called `codeFormatter` instead ?

src/components/otp-field/OTPClock.tsx
Domino9697: Yes I did but the thing is that we are rerendering the component each time so we will always execute the function. Using setInterval changes absolutely nothing :)

src/components/otp-field/OTPClock.tsx
Domino9697: Well you never generate an OTP code in leeloo or in the Popup so, you will have lost some time between the actual generation and the display. SO I would argue we leave it as mandatory.

src/components/otp-field/OTPClock.tsx
Domino9697: timeRef is a reference created by react and as such is never null. However, its inner value (timeRef.current) can be null and we check if it is not null here :)

src/components/otp-field/helpers.ts
Domino9697: I am using this helper file for more than just codeFormatter. It also is used to compute the perimiter of the filled path. But this is a good point. 

We rely a lot on helper files. But then to me, within the files, you will just find some functions that you don't need to be within a react component.

src/components/otp-field/OTPField.spec.tsx
```diff
@@ -0,0 +1,141 @@
+import * as React from 'react';
+import { mount } from 'enzyme';
```
Aktanee, line 2: Out of scope of this MR: Since we moved to RTL in `leeloo`, does it make sense to someday move toward this in `ui-components`?

src/components/otp-field/OTPField.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+import 'jest-styled-components';
+import { OTPField } from '@components/otp-field/OTPField';
+
+Date.now = jest.fn().mockReturnValue(1592829201510);
+
+const DEFAULT_LABEL = 'security code';
+const DEFAULT_CODE = '000000';
+const CURRENT_DATE = Date.now();
+const VALIDITY_PERIOD = 30000;
```
Aktanee, line 11: Is the `VALIDITY_PERIOD` a fix value?

src/components/otp-field/OTPField.spec.tsx
```diff
@@ -0,0 +1,141 @@
...
+    });
+  });
+
+  describe('OTP Field code shuffler', () => {
+    it('should trigger a code shuffle when there is no time left', () => {
+      const otpField = mount(
+        <OTPField
+          code={DEFAULT_CODE}
+          label={DEFAULT_LABEL}
+          validityEndDate={CURRENT_DATE - 1000}
```
Aktanee, line 105: Since you are using multiple times the pattern `CURRENT_DATE - 1000` maybe you can add it as a variable `VALIDITY_ENDED`?

src/components/otp-field/OTPField.mdx
```diff
@@ -0,0 +1,101 @@
...
+    label={'security code'}
+    code={'145236'}
+    validityPeriod={60000}
+    validityEndDate={Date.now() + 20000}
+  />
+</Playground>
+
+### Alert Threshold
+
+Use the `alertThreshold` prop to change the color to an alert color when the the threshold is reached.
```
Aktanee, line 75: Will we have a verification that the threshold isn't superior than the `validityPeriod`?

Domino9697: If it is, no problem, the entire path will be red :) 
So no need to check this IMO !

src/components/otp-field/OTPClock.tsx
pducolin: So when initializing `timeRef` with `useRef(null)`, it's `timeRef.current` that is set to `null` ?

src/components/otp-field/OTPClock.tsx
pducolin: I'm not 100% convinced of using `useEffect` and `setTimeout`. Let's discuss tomorrow

src/components/otp-field/OTPClock.tsx
Domino9697: Yes exactly !

pducolin: Ah! TIL useRef!

src/components/otp-field/helpers.ts
oclero: My opinion about it: if this helper functions are just for OTPField, and you don't have a lot of functions that you could spread into multiple helper files, sorted by kind, a single helper file in the folder is fine.

src/components/otp-field/OTPField.spec.tsx
oclero: Could you add a comment about this. Is it days? milliseconds?

src/components/otp-field/OTPClock.module.scss
```diff
@@ -0,0 +1,42 @@
...
+  stroke-width: 2px;
+
+  transform: rotate(90deg);
+  transform-origin: center;
+
+  transition: stroke 250ms ease-out;
+
+  transition: stroke-dasharray 1s linear;
+
+  stroke: colors.$mid-green-00;
```
oclero, line 33: Why all these empty lines?

src/components/otp-field/OTPField.spec.tsx
oclero: I agree!

src/components/otp-field/OTPField.tsx
```diff
@@ -0,0 +1,50 @@
...
+import styles from './OTPField.module.scss';
+
+export interface OTPFieldProps extends OTPClockProps {
+  label: string;
+  /**
+   * A 6-character otp code string
+   * No space character should be within the code
+   * If null is provided, the component will shuffle the code
+   */
+  code: string | null;
```
oclero, line 15: Why not `code?: string;` ?

src/components/otp-field/OTPField.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+  React.useEffect(() => {
+    setIsShuffling(!code);
+  }, [code]);
+
+  const { shuffledCode } = useOTPCodeShuffler({
+    shouldShuffle: isShuffling
+  });
+
+  const formattedCode = formatCode(isShuffling || !code ? shuffledCode : code);
```
oclero, line 39: can't you do something like this? `code ?? shuffleCode`

src/components/otp-field/helpers.ts
```diff
@@ -0,0 +1,29 @@
...
+  return [code.slice(0, 3), ' ', code.slice(3)].join('');
+}
+
+const ANIMATION_TIME = 1000;
```
oclero, line 5: Could you add the unit in commentary at the end of line: `// ms`

src/components/otp-field/helpers.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  const adaptedTimeFraction =
+    timeFraction - (ANIMATION_TIME / validityPeriod) * (1 - timeFraction);
+
+  const circleDashArrayValue = adaptedTimeFraction * perimiter;
+
+  const positiveCircleDashArrayValue =
+    circleDashArrayValue > 0 ? circleDashArrayValue : 0;
+
+  return positiveCircleDashArrayValue.toFixed();
```
oclero, line 28: I would keep the returned value as a number, and delay the transformation to string at the last moment. This makes this function dealing only with numbers, as it should do, and delegating the formatting to the user of the func.

src/components/otp-field/OTPField.spec.tsx
Domino9697: Yes I agree as well !

src/components/otp-field/OTPField.spec.tsx
Domino9697: I will add this in the actual code not the tests

src/components/otp-field/OTPField.tsx
Domino9697: Because I want to force the user to use this prop. 
Basically, you will pass a value that is either null or a string code.

PR #129, opened 2020-06-23
Feat/add range slider component

pducolin:
Slider based on input type range
WIP as I'm adding tests.

Potential controversial points:
* tooltip positioning
* Trackbar color evaluation

image


src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,219 @@
...
+      to right,
+      var(--track-fill-color) var(--fill-percentage-value),
+      var(--track-background-color) var(--fill-percentage-value)
+    );
+
+    /*
+    * Removes the focus border.
+    */
+    &:focus {
+      outline: none;
```
sericaia, line 76: what is the impact of this line in accessibility?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## [3.0.1](https://github.com/Dashlane/ui-components/compare/v3.0.0...v3.0.1) (2020-06-17)
```
oclero, line 1: Why is this file part of the pull request?

src/components/Slider/Slider.module.scss
pducolin: focus is visible through the thumbnail background color, even if it is less clear than the outline. will check with designers for improving this

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@dashlane/ui-components",
-  "version": "3.0.0",
+  "version": "3.0.1",
```
oclero, line 3: Same question here :)

CHANGELOG.md
pducolin: it's part of the automated commit, don't know :(

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,132 @@
...
+      return;
+    }
+
+    const range = max - min;
+    const valuePercentage = ((value - min) / range) * 100;
+    slider.style.setProperty('--fill-percentage-value', `${valuePercentage}%`);
+  }, [sliderRef, value, max, min]);
+
+  const handleChangeDone = () => {
+    onChangeComplete && onChangeComplete(value);
```
sericaia, line 46: * add `return`
* prefer  `onChangeComplete ? onChangeComplete(value) : null` instead

Question:
* why do you need to send the value when it is complete?

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,132 @@
...
+  };
+
+  React.useEffect(() => {
+    const slider = sliderRef.current;
```
sericaia, line 50: `const { current: slider } = sliderRef;` ?

CHANGELOG.md
pducolin: good catch! fixed with a rebase

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,132 @@
...
+  }, [sliderRef, value, handleChangeDone]);
+
+  React.useEffect(() => {
+    const { current: slider } = sliderRef;
+    const { current: tooltipElement } = tooltipRef;
+    if (!slider || !tooltipElement) {
+      return;
+    }
+
+    const halfThumbWidth =
```
sericaia, line 74: can this logic be moved to a function outside this component and properly tested? 🙏

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,132 @@
...
+        name={`slider-${label}`}
+        type='range'
+        ref={sliderRef}
+        min={min}
+        value={value}
+        max={max}
+        step={step}
+        className={styles.sliderRange}
+        onChange={event => {
+          const floatValue = parseFloat(event.currentTarget.value);
```
sericaia, line 113: maybe move these lines to a function outside the input and call it here?

src/components/Slider/Slider.tsx
pducolin: It's to keep a consistent behavior with legacy slider - based on `react-slider`. We use it in the password generator, and trigger a new password generation `onChangeComplete`.

What about: `onChangeComplete ?? onChangeComplete(value)` ?

pducolin: As I didn't say it, google told me :)

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  React.useEffect(() => {
+    const slider = sliderRef.current;
+
+    if (!slider) {
+      return;
+    }
+
+    const range = max - min;
+    const valuePercentage = ((value - min) / range) * 100;
```
oclero, line 41: Check if `range !== 0`

src/components/Slider/Slider.tsx
sericaia: ok, keep it consistent with the value.

sericaia:
Is it possible to send a custom `aria-labelledby` value and define the label outside?

src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const centerPosition = slider.clientWidth / 2;
+
+    const percentOfRange = (value - min) / (max - min);
+    const valuePxPosition = percentOfRange * slider.clientWidth;
+    const distFromCenter = valuePxPosition - centerPosition;
+    const percentDistFromCenter = distFromCenter / centerPosition;
+
+    const offset = percentDistFromCenter * halfThumbWidth;
+
+    const tooltipPosition = valuePxPosition - halfTooltipWidth - offset;
```
oclero, line 92: As Daniela says, maybe all these computation could be moved to another file. For this kind of code, I like to have a `SliderGeometry` type that contains all the necessary information (and can be tested idependently).

src/components/Slider/Slider.tsx
oclero: With useCallback?

src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,219 @@
...
+    &:hover,
+    &:focus {
+      &::-webkit-slider-thumb {
+        background-color: var(--thumb-hover-color);
+      }
+      &::-moz-range-thumb {
+        background-color: var(--thumb-hover-color);
+      }
+      &::-ms-thumb {
+        background-color: var(--thumb-hover-color);
```
oclero, line 114: Could you group these 3 rules together?

src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,219 @@
...
+    &:active {
+      &::-webkit-slider-thumb {
+        background: var(--thumb-active-color);
+      }
+      &::-moz-range-thumb {
+        background: var(--thumb-active-color);
+      }
+      &::-ms-thumb {
+        background: var(--thumb-active-color);
+      }
```
oclero, line 127: Also here: you can group.
Btw, is there a standard `::thumb {}` pseudo-element ?

src/components/Slider/Slider.module.scss
```diff
@@ -0,0 +1,219 @@
...
+        background-color: var(--disabled-color);
+      }
+      &::-moz-range-thumb {
+        cursor: default;
+        background-color: var(--disabled-color);
+      }
+      &::-ms-thumb {
+        cursor: default;
+        background-color: var(--disabled-color);
+      }
```
oclero, line 154: Please group these rules.

pducolin: I tried but it didn't work :(

pducolin: I tried grouping with `&::-webkit-slider-thumb, &::-moz-range-thumb {}` and it didn't work, I'm probably doing something wrong

pducolin:
> Is it possible to send a custom `aria-labelledby` value and define the label outside?

You mean take the label out of the component? I discussed about it with design, and changed my mind on it a couple of times. We decided label should be part of the component - as it shows current value too. What do you think?

src/components/Slider/Slider.module.scss
oclero: Ha? Really strange, indeed.

src/components/Slider/Slider.module.scss
pducolin: no standard element, more infos https://css-tricks.com/styling-cross-browser-compatible-range-inputs-css/

Most mindblowing thing I found out is that IE provides with property to set different colors for left and right side of the bar, and Chrome does not

oclero: Ah ok, it's not standardized yet, unfortunately.

sericaia:
> > Is it possible to send a custom `aria-labelledby` value and define the label outside?
> 
> You mean take the label out of the component? I discussed about it with design, and changed my mind on it a couple of times. We decided label should be part of the component - as it shows current value too. What do you think?

Some libs give that possibility and I think it is good ;)

src/components/Slider/Slider.module.scss
pducolin: Found why: https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list
```a single unsupported selector in a selector list invalidates the whole rule.```

File: src/components/slider/Slider.module.scss [hunk 13]
```diff
@@ -0,0 +1,235 @@
...
+
+    /* All the same for Firefox and IE */
+    // Cannot group these selectors within a selector list because
+    // a single unsupported selector in a selector list invalidates the whole rule.
+    // from: https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list
+    &::-moz-range-thumb {
+      @include thumb;
+    }
+    &::-ms-thumb {
+      @include thumb;
```
oclero, line 98: Cool :)

src/components/slider/Slider.spec.tsx
```diff
@@ -0,0 +1,41 @@
...
+describe('<Slider />', () => {
+  it('should render default', () => {
+    const label = 'Slider';
+    const value = 50;
+    const { asFragment, queryByText } = render(
+      <Slider value={value} label={label} />
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+    expect(queryByText(`${label} (${value})`)).not.toBe(null);
```
Domino9697, line 15: Should we put not null ? Or toBeInTheDocument ?

src/components/slider/geometry/SliderGeometry.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export const evaluateSliderTooltipLeftPosition = ({
+  thumbWidth,
+  sliderWidth,
+  tooltipWidth,
+  valuePercentage
+}: SliderTooltipPositionProps) => {
+  const halfThumbWidth = thumbWidth / 2;
+  const halfTooltipWidth = tooltipWidth / 2 - 2;
```
Domino9697, line 27: Why - 2 ? Could you add a comment ?

src/components/slider/Slider.spec.tsx
pducolin: ah true! I always forget about that

src/components/Slider/Slider.tsx
Domino9697: Javascript ... :rofl:

src/components/slider/geometry/SliderGeometry.ts
pducolin: Truth is I tried to find a reason why, and I could not find why there's this 2px offset

src/components/Slider/Slider.tsx
```diff
@@ -37,17 +42,18 @@ export const Slider = ({
...
+      '--fill-percentage-value',
+      `${evaluateSliderValuePercentage(value, min, max) * 100}%`
+    );
   }, [sliderRef, value, max, min]);
 
-  const handleChangeDone = () => {
-    onChangeComplete && onChangeComplete(value);
-  };
+  const handleChangeDone = React.useCallback(() => {
+    onChangeComplete ? onChangeComplete(value) : null;
```
sericaia: `return` is missing.

_Non-blocking note_: I actually prefer the syntax:

```
if(onChangeComplete) {
return onChangeComplete(value);
}
```

src/components/slider/Slider.spec.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { fireEvent, render } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
```
sericaia, line 3: can we have this globaly to avoid adding it in every file?

src/components/Slider/Slider.tsx
pducolin: I tend to avoid returning in void functions, except for early returns. Is it usual for `useCallback` hook to always use return ?

PR #130, opened 2020-06-23
Adding a flexible infoBox 

MatMath:
Adding a **InfoBox** to the UI components following the color theme.

### Feature: (see mdx file)
Color Scheme(4): subtle / strong / warning / alert
Size options(3): small / simple / descriptive
Actions: Nothing or Secondary &/or Primary Button
Coverage: 100%

Screen Shot 2020-06-23 at 4 46 46 PM
Screen Shot 2020-06-23 at 4 46 36 PM
Screen Shot 2020-06-23 at 4 47 19 PM

Missing (future PR): The Warning & Alert severity should have different Button color following the theme, but the current Button only support light/dark theme so I felt it would be out of the scope of this PR to change the button theme while adding a new feature.


oclero:
Hey, looks like the generated docs auto-commit is missing.

src/components/info-box/types.ts
```diff
@@ -0,0 +1,12 @@
...
+  STRONG: 'strong',
+  SUBTLE: 'subtle',
+  WARNING: 'warning'
+} as const;
+
+export const InfoBoxSize = {
+  SMALL: 'small',
+  SIMPLE: 'simple',
+  DESCRIPTIVE: 'descriptive'
+} as const;
```
oclero, line 12: Could you follow the same convention as for the other components?

```ts
export type InfoBoxSeverity = 'alert' | 'strong' | 'subtle' | 'warning';

export type InfoBoxSize = 'small' | 'simple' | 'descriptive';
```

src/components/info-box/InfoBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { InfoCircleIcon } from '@components/icons';
+import { Button } from '@components/button/Button';
+import { dashGreen00, functionalRed00, orange00, white } from '@colors';
+import { getClassNames } from '@utility/cssUtils';
+
+import { InfoBoxSeverity, InfoBoxSize } from './types';
+import styles from './InfoBox.module.scss';
+
+type InfoBoxSizeList = 'small' | 'simple' | 'descriptive'; // To display via mdx
+type InfoBoxSeverityList = 'alert' | 'strong' | 'subtle' | 'warning'; // To display via mdx
```
oclero, line 12: See my other comment about these types. No need to declare them with two different ways, it will be harder to maintain. Please prefer the convention we chose for the other components.

src/components/info-box/InfoBox.tsx
```diff
@@ -0,0 +1,93 @@
...
+      case InfoBoxSeverity.WARNING:
+        return orange00;
+      case InfoBoxSeverity.SUBTLE:
+      default:
+        return dashGreen00;
+    }
+  };
+
+  const showActions = () => {
+    return (
```
oclero, line 50: Could you move this function out of the functional component, so we avoid a function creation every time?

src/components/info-box/InfoBox.tsx
MatMath: Hi @oclero 
The UI-Component does have 2 different ways, ex `<Button>` for list and `<Alert>` for enum typing. 

If we define the component as here `type "a" | "b" | ...` the Gatsby docs show properly 👍. But we lose the flexibility of typescript with `InfoBoxSeverity.WARNING` and we have to hardcode the value in the code logic itself. It was the closes I came to bridge the Docs & TS flexibility but I understand and I will remove the const typing.

oclero: Yes I understand but we don't use enums in ui-componets, unfortunately. You still get autocompletion and typing: it's not as bad as hardcoded strings.

MatMath:
> Hey, looks like the generated docs auto-commit is missing.

@oclero I saw a creeping commit few days ago `docs: update docz documentation` but that includes 120+ files so I removed it. I thought that would be done in a CI-CD chore on merge, not part of a commit since it would create merging conflict hell and render the review impossible to do. It is because most of the changes were made on filepath!?! 
ie: `"fullpath":"/Users/pducolin/repos/ui-components/src/components/sidenav/Sidenav.mdx",`
to: `"fullpath":"/Users/mlegault/github/ui-components/src/components/sidenav/Sidenav.mdx",`
All `filename Hash` also changed due to the `filepath`...

If I manually ran `yarn exec-docz-update` nothing gets added. I found how I can trick it to trigger but it is still `68 modified and 79 added`. I can add them but I think it should be part of the CI-CD instead.

src/components/info-box/InfoBox.tsx
```diff
@@ -0,0 +1,104 @@
...
+}: Partial<InfoBoxProps>) => {
+  return (
+    <div className={styles.actionBox}>
+      {secondary ? (
+        <Button
+          nature='secondary'
+          size='small'
+          theme={severity === 'strong' ? 'dark' : 'light'}
+          className={styles.actionsBtn}
+          onClick={() => secondary()}
```
sanatshah, line 39: Nitpick : You could just pass the function itselft -> onClick={secondary}

sanatshah:
Your commits seem to be unverified.  Ran into a similar thing before, had to add another gpg key for github and update my local git config for this repo to use that key

src/components/info-box/InfoBox.tsx
MatMath: I wish but lint does not agree with me. :( 
onClick is not the same type

sanatshah: Oh weird, sometime lint has a mind of its own

PR #132, opened 2020-06-25
Docs/update docs with rtl

pducolin:
Updating docs with React Testing Library migration. I also updated Button's tests - as they are used as an example in docs


documentation/COMPONENT_STEPBYSTEPGUIDE.md
```diff
@@ -171,25 +170,25 @@ describe('<Button>', () => {
 
   describe('Global render', () => {
     it('should render default', () => {
-      wrapper = createWrapper({});
-      expect(wrapper).toMatchSnapshot();
-      expect(wrapper.exists('.medium')).toEqual(true);
+      const { asFragment, container } = createWrapper({});
```
adelego, line 173: We need to change the test here. If we write tests the same way we do with enzyme, we are missing the point of switching libraries.
In my opinion, it is critical to test the ui in this lib as it is what we mostly provide. But RTL is more adapted to testing ux.
To me we need to clarify our testing strategy. I propose:
- keeping snapshots for ui as a short term solution. Switching to real tool to prevent UI regression when possible
- using rtl to test UX and prevent it from regressing

documentation/COMPONENT_STEPBYSTEPGUIDE.md
```diff
@@ -150,19 +150,18 @@ Now you [can launch docz](../CONTRIBUTING.md#developing-on-docz) and see your wo
 
 > 🐞docz has a known bug with props cache and TypeScript, sometimes Props aren't displayed in the documentation. docz tasks (`yarn docz:dev` and `yarn docz:build`) will clean up docz cache automatically, but if you experience this while docz is runing in watch mode, kill it and relaunch it
 
-7. Last but not least, let's add some tests. We use [jest](https://jestjs.io/) and [Enzyme](https://enzymejs.github.io/enzyme/) to run snapshot and unit tests. Create a file `Button.spec.tsx`
```
lhoursquentin, line 153: Maybe we should wait until all tests are using react testing library before removing completely `enzyme` from the guide, this might be confusing for people looking at other tests. Just a mention to enzyme would be good enough.

pducolin: I see your point. We decided to remove also `styled-component` from documentation, even if some components are still using it. Goal of documentation is to be in line with current tech stack. I can add a note saying that enzyme is obsolete.

documentation/COMPONENT_STEPBYSTEPGUIDE.md
pducolin: Would you use something else to generate snapshots ?

adelego: I don't really mind if we use RTL or enzyme to create snapshots.
But this doc shows confusion on how we want to test against regression in our lib, why we are migrating to RTL. There is no point to migrating if we write the exact same tests

documentation/COMPONENT_STEPBYSTEPGUIDE.md
lhoursquentin: Sure a note noting that it is obsolete is fine with me :slightly_smiling_face:

documentation/COMPONENT_STEPBYSTEPGUIDE.md
sericaia: I agree with @adelego, this is not the right way to use RTL ;)


Can you check the following resources?
* https://testing-library.com/docs/guide-which-query
* https://testing-library.com/docs/dom-testing-library/api-queries
* https://www.robinwieruch.de/react-testing-library
* https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

src/components/button/Button.spec.tsx
```diff
@@ -1,112 +1,112 @@
...
+      const text = 'A button';
+      const { getByText } = render(
+        <Button onClick={() => window.alert('hello')} disabled>
+          {text}
+        </Button>
       );
 
       window.alert = jest.fn();
-      button.simulate('click');
+      fireEvent.click(getByText(text));
```
sericaia, line 109: can we use `userEvents` instead of `fireEvent`?

documentation/COMPONENT_STEPBYSTEPGUIDE.md
pducolin: As discussed privately with @adelego I will update the guide to add tests for components UX (for example, click on button should trigger the relative event)

src/components/button/Button.spec.tsx
pducolin: yes, will check that

pducolin: too bad they don't have a press or drag event (yet) https://github.com/testing-library/user-event

src/components/button/Button.spec.tsx
```diff
@@ -1,112 +1,112 @@
...
-      const button = mount(
+      const { asFragment, queryByText } = render(
         <Button>
           <label>{text}</label>
         </Button>
       );
 
-      expect(button.find('label')).toBeTruthy();
+      expect(asFragment()).toMatchSnapshot();
+      expect(queryByText(text)).toBeInTheDocument();
```
adelego, line 84: Maybe `getByRole('label, {name: text}`?

pducolin: label has no aria-role :( 
https://w3c.github.io/html-aria/

sericaia: cant we use this?


* ByLabelText find by label or aria-label text content
```
getByLabelText
queryByLabelText
getAllByLabelText
queryAllByLabelText
findByLabelText
findAllByLabelText
```
(from https://testing-library.com/docs/react-testing-library/cheatsheet)

pducolin: Not working, I think `LabelText` finds the item that is labelled by a `label` with some `text` in it. Here the button contains a `label`, that is not labelling it.

src/components/button/Button.spec.tsx
sericaia: yet!

documentation/COMPONENT_STEPBYSTEPGUIDE.md
```diff
@@ -150,46 +150,65 @@ Now you [can launch docz](../CONTRIBUTING.md#developing-on-docz) and see your wo
...
       });
-      expect(wrapper).toMatchSnapshot();
-      expect(wrapper.exists('.large')).toEqual(true);
+      expect(asFragment()).toMatchSnapshot();
+    });
+  });
+
+  describe('Interaction', () => {
+    it('should execute action on click', () => {
+      const { getByRole } = render(
```
sericaia, line 194: replace by screen ;)

documentation/COMPONENT_STEPBYSTEPGUIDE.md
```diff
@@ -150,46 +150,65 @@ Now you [can launch docz](../CONTRIBUTING.md#developing-on-docz) and see your wo
...
 import * as React from 'react';
-import { mount, ReactWrapper } from 'enzyme';
-import 'jest-styled-components';
+import { render } from '@testing-library/react';
 import { Button, ButtonProps } from './Button';
 
 describe('<Button>', () => {
   let wrapper: ReactWrapper;
 
   const createWrapper = ({ nature, size, children }: ButtonProps) => {
```
sericaia, line 163: nonblocking: maybe wrapper is no longer the best name. `renderButton`?

documentation/COMPONENT_STEPBYSTEPGUIDE.md
sericaia: it was not fixed, right?

src/components/button/__snapshots__/Button.spec.tsx.snap
```diff
@@ -1,80 +1,95 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`<Button> Children render should render the specified children 1`] = `
+<DocumentFragment>
```
sericaia, line 4: I wonder why you used the `asFragment` instead of the `container`. With `asFragment` we get a lot of these `<DocumentFragment>`

pducolin: Source is https://www.leighhalliday.com/introduction-react-testing-library
Will change it

documentation/COMPONENT_STEPBYSTEPGUIDE.md
pducolin: forgot this one, good catch

documentation/COMPONENT_STEPBYSTEPGUIDE.md
```diff
@@ -150,46 +150,59 @@ Now you [can launch docz](../CONTRIBUTING.md#developing-on-docz) and see your wo
...
     });
 
     it('should render large', () => {
-      wrapper = createWrapper({
+      const { content } = renderButton({
         size: 'large'
       });
-      expect(wrapper).toMatchSnapshot();
-      expect(wrapper.exists('.large')).toEqual(true);
+      expect(asFragment()).toMatchSnapshot();
```
sericaia, line 186: ? something is wrong here :P

pducolin: ah, not having linter in mdx is a pain

pducolin: should be good now! (copy pasted test from Button)

PR #133, opened 2020-06-25
Fix/slider remove label

pducolin:
Some fixes to the sliders:
* removed the label, now there are an id property and a labelledBy aria property
* centered value in the tootlip 

Before:
image

After:
image


src/components/slider/__snapshots__/Slider.spec.tsx.snap
```diff
@@ -5,15 +5,13 @@ exports[`<Slider /> should render default 1`] = `
...
-      Slider (50)
-    </label>
     <input
+      aria-labelledby="Slider"
+      aria-valuemax="100"
+      aria-valuemin="0"
+      aria-valuenow="50"
       class="sliderRange"
-      id="slider-Slider"
+      id="Slider"
```
sericaia, line 14: why is the id capitalized?

pducolin: ah, I forgot to change the string I was using for the label

src/components/slider/Slider.tsx
```diff
@@ -8,7 +8,8 @@ import {
 } from './geometry/SliderGeometry';
 
 export interface SliderProps {
-  label: string;
+  id: string;
+  ariaLabelledby: string;
```
oclero, line 12: Why have you renamed the prop? You could have kept the same name, but just used it in `aria-labelledby`. You kept `min` and `max` (and not `aria-min` and `aria-max`) so it's better to keep the same naming convention for all props, isn'it ?

pducolin: `min` and `max` are also used for the input's `min|max` values. Here I preferred to be explicit, as I considered that using `label` could have made the user think of a visible attribute. What do you think?

a-coruble:
Could you add some screenshots of before/after ? Thanks for the fix !

src/components/slider/Slider.spec.tsx
```diff
@@ -8,33 +8,32 @@ describe('<Slider />', () => {
     const label = 'Slider';
     const value = 50;
     const { asFragment, queryByText } = render(
-      <Slider value={value} label={label} />
+      <Slider value={value} ariaLabelledby={label} id={label} />
```
adelego, line 11: LabelledBy implies it expects an id to an other element, not a text. It doesn't look like this is the case here.

adelego: You are not using the label so this works in any case. But this is misleading as to how to use your slider and what its props are.

pducolin: It is indeed expecting an id from another element. In the test I didn't add a separate element but I could.

pducolin: I could skip the id - and use the label passed in props. I need the id for the output

pducolin: I removed the id and added a comment on the ariaLabelledBy prop, do you think it is still misleading? Another solution would be to pass a ref to an element

src/components/slider/Slider.tsx
fmatias-dashlane: This is a breaking change, the commit should be marked as such to avoid unexpected issues on code using this component 🙏

pducolin: there's no code currently using it, but I see your point, changing it

pducolin: this PR is more a fix on previous PR that added this component

src/components/slider/Slider.spec.tsx
a-coruble: I prefer your solution with the id

src/components/slider/Slider.tsx
fmatias-dashlane: yeah, I figured this might not be in use yet but that shouldn't matter from this project's "point of view" 😅  Thanks 🙏

fmatias-dashlane: yeah, I figured this might not be in use yet but that shouldn't matter from this project's "point of view" 😅
Thanks 🙏

src/components/slider/Slider.spec.tsx
adelego: I like the solution with id as well, I just think we should change the test to it clear how to use it. You could do something like
```
render(
     <>
          <label id="label-id">This is a label</label>
          <Slider value={value} ariaLabelledby="label-id" id={label} />
     </>

```
Or at the very least, rename the `label` variable `labelId`

adelego: Actually why is the slider labelled by itself. If the label is internal, it shouldn't be a prop

pducolin: The label is external as per this PR, discussion ongoing

pducolin: I'm adding a label in the documentation, live docs should be used to see how to use a component rather than tests. What do you think?

src/components/slider/Slider.spec.tsx
```diff
@@ -5,36 +5,35 @@ import { Slider } from './Slider';
...
     expect(queryByText('100')).toBeInTheDocument();
   });
 
   it('should render with a tooltip on press', () => {
     const value = 30;
-    const label = 'Slider';
-    const { getByLabelText, queryByText } = render(
-      <Slider value={value} label={label} />
+    const label = 'slider';
+    const { getByRole, queryByText } = render(
```
sericaia, line 22: you dont need this ;)

you can just do 

`const slider = screen.getByRole('slider');`

sericaia: fyi https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

pducolin: cool! thank you

pducolin: changed

src/components/slider/Slider.spec.tsx
```diff
@@ -1,40 +1,37 @@
...
   it('should render default', () => {
     const label = 'slider';
     const value = 50;
-    const { asFragment, queryByText } = render(
-      <Slider value={value} ariaLabelledby={label} id={label} />
+    const { asFragment } = render(
+      <Slider value={value} ariaLabelledby={label} />
     );
 
     expect(asFragment()).toMatchSnapshot();
```
sericaia, line 14: cant we use screen.asFragment too?

src/components/slider/Slider.spec.tsx
```diff
@@ -1,40 +1,37 @@
...
     const value = 30;
     const label = 'slider';
-    const { getByRole, queryByText } = render(
-      <Slider value={value} ariaLabelledby={label} tooltip={false} id={label} />
-    );
-    const slider = getByRole('slider');
+    render(<Slider value={value} ariaLabelledby={label} tooltip={false} />);
+    const slider = screen.getByRole('slider');
+    // simulate mouse pressed event
     fireEvent.mouseDown(slider);
```
sericaia, line 34: can we add `userEvent` in this project? :(

src/components/slider/Slider.spec.tsx
pducolin: no `asFragment` in the screen object. Don't know if they have another way to do snapshots

src/components/slider/Slider.spec.tsx
pducolin: I knew you would have asked! 😄  I promise I'll update this in the other PR, rebasing it, after merging this one.

pducolin: Can't use `userEvent` here as there's no `mousePressed` event on `userEvent`

PR #135, opened 2020-06-25
Feat/update textinput components

oclero:
`TextInput` and `PasswordInput` were not up-to-date to Figma latest revision. Moreover, they contained some CSS issues. Now it's fixed.

#### `TextInput`

- Fix read-only state when combined with a `FeedbackType`.
- Fix disabled state when combined with a `FeedbackType`.
- Add `label` prop.
- Update docs.
- Update test snapshots.

**Before:**
Screenshot_2020-06-26 TextInput(1)

**After:**

Screenshot_2020-06-26 TextInput(2)

#### `PasswordInput`

- Fix wrong line-height for label (was 20px, should be 16px).
- Hide Edge's default Eye icon.
- Fix margin between text and hide/reveal button. It was noticeable when the password was overflowing.
   
  **After:**
Screenshot_2020-06-26 Password Input(3)

- Fix disabled state: hide/reveal button should be disabled too (unlike read-only state).
   **Disabled:**
   Screenshot_2020-06-26 Password Input

   **Read-only:**
   Screenshot_2020-06-26 Password Input(1)

- Update docs.
- Update test snapshots.

## Left to do

Since I'm leaving Dashlane today, here is what's left to do:
 - Re-create the commits, with a `BREAKING CHANGE` label.
 - Re-write the tests with RTL. The current tests are not good at all and don't really test anything.
 - Update Leeloo's dependency of UI-Components so Leeloo gets the latest changes.
 - Replace text inputs and password inputs in Leeloo with the components from UI-Components.

@a-coruble Since this PR goal was to help you, you're the best to handle it now. It's my farewell gift ;)


adelego:
Could you add before/after screenshots please?

src/components/password-input/PasswordInput.module.scss
```diff
@@ -10,88 +11,109 @@
   display: flex;
   flex-direction: column;
   align-items: flex-start;
+}
 
-  .label {
-    @include typography.body-small;
+.label {
+  @include typography.body-small;
+  overflow-wrap: anywhere;
```
ni0ki, line 18: The `anywhere` property is not supported on Safari

src/components/text-input/TextInput.module.scss
```diff
@@ -17,6 +18,17 @@
   }
 }
 
+.label {
+  @include typography.body-small;
+  overflow-wrap: anywhere;
```
ni0ki, line 23: The `anywhere` property is not supported on Safari

src/components/password-input/PasswordInput.module.scss
oclero: Thanks! I will try a workaround

src/components/password-input/PasswordInput.module.scss
```diff
@@ -10,88 +11,109 @@
...
-    text-transform: uppercase;
-    height: 20px;
-    font-size: 10px;
-    line-height: 20px;
-    font-weight: 600;
-    margin-bottom: 4px;
-  }
+  color: colors.$dash-green-01;
+  text-transform: uppercase;
+  font-size: 10px;
```
adelego, line 21: Why are you completely overriding the typo properties?
10px doesn't seem standard. Font sizes are supposed to be multiples of 4.

src/components/password-input/PasswordInput.tsx
```diff
@@ -80,7 +78,8 @@ export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
           className={getClassNames(styles.input, styledClassName)}
         />
         <button
-          className={getClassNames(styles.icon)}
+          disabled={prop.disabled}
+          className={getClassNames(styles.revealButton)}
```
adelego: No need for classNames here

src/components/text-input/TextInput.module.scss
```diff
@@ -43,31 +55,43 @@
...
+  &:read-only.error,
+  &:read-only ~ .inlineText,
+  &:focus:read-only ~ .inlineText {
     background: colors.$dash-green-06;
     border-color: colors.$dash-green-05;
   }
 
-  &:disabled {
+  &:disabled,
+  &:disabled ~ .inlineText {
```
adelego, line 84: I don't know this selector ~. What does it do?

src/components/password-input/PasswordInput.module.scss
pducolin: Question is why is this still here. As I removed the label, it should be removed

src/components/text-input/TextInput.tsx
```diff
@@ -8,47 +8,117 @@ type _ForbiddenProps = 'size' | 'prefix' | 'type';
...
+ * @param props
+ */
+const InlineText: React.FC<{
+  disabled?: boolean;
+  inlineText?: string;
+  inlineCss?: string;
+}> = ({ disabled, inlineText, inlineCss }) => {
+  return (
+    <span
+      className={getClassNames(styles.inlineText, inlineCss, {
```
adelego, line 44: `inlineCSS` is not a very good naming in my opinion, this should contain `className`. Maybe `inlineCustomClassName`?

src/components/password-input/PasswordInput.module.scss
oclero: @adelego See "Web Kit" on Figma
@pducolin I didn't know that. Why is there still a label in PasswordInput then?

src/components/text-input/TextInput.tsx
adelego: Also shouldn't it be last in the list so a user can override the default style?

src/components/text-input/TextInput.tsx
```diff
@@ -8,47 +8,117 @@ type _ForbiddenProps = 'size' | 'prefix' | 'type';
...
+ * Feedback displayed below.
+ * @param props
+ */
+const Feedback: React.FC<{
+  feedbackText?: string;
+  feedbackType?: TextInputFeedbackType;
+  fullWidth?: boolean;
+  disabled?: boolean;
+}> = ({ feedbackText, feedbackType, fullWidth, disabled }) => {
+  const feedbackCssClass =
```
adelego, line 106: `feedbackClassName` is more coherent IMO

src/components/password-input/PasswordInput.module.scss
adelego: @oclero ok but does that mean we need to update typos? Is there a new standard for all typos? just for inputs? Are we not following the x4 rules anymore? Can you ask a designer please?

src/components/password-input/PasswordInput.tsx
```diff
@@ -41,6 +41,61 @@ export interface PasswordInputProps
   showPasswordTooltipText: string;
 }
 
+/**
+ * Label displayed above.
+ * @param props
+ */
+const Label: React.FC<{ label?: string }> = ({ label }) => {
+  return label ? (
+    <span className={getClassNames(styles.label)}>{label}</span>
```
adelego, line 50: not need for `getClassNames` here

src/components/password-input/PasswordInput.tsx
```diff
@@ -41,6 +41,61 @@ export interface PasswordInputProps
...
+  onClick,
+  hidePasswordTooltipText,
+  showPasswordTooltipText
+}) => {
+  const RevealOrHideIcon = showPassword ? HideIcon : RevealIcon;
+
+  return (
+    <button
+      disabled={disabled}
+      className={getClassNames(styles.revealButton)}
```
adelego, line 76: Same here

src/components/password-input/PasswordInput.tsx
```diff
@@ -41,6 +41,61 @@ export interface PasswordInputProps
...
+};
+
+/**
+ * Button to reveal/obfuscate the password.
+ * @param props
+ */
+const RevealButton: React.FC<{
+  showPassword?: boolean;
+  disabled?: boolean;
+  onClick?: () => void;
```
adelego, line 61: This does not match a regular `onClick`type signature. It should take a click event as a parameter.

src/components/password-input/PasswordInput.tsx
```diff
@@ -41,6 +41,61 @@ export interface PasswordInputProps
...
+  hidePasswordTooltipText: string;
+  showPasswordTooltipText: string;
+}> = ({
+  showPassword,
+  disabled,
+  onClick,
+  hidePasswordTooltipText,
+  showPasswordTooltipText
+}) => {
+  const RevealOrHideIcon = showPassword ? HideIcon : RevealIcon;
```
adelego, line 71: You should have useMemo here instead of unnecessarily re-rendering this icon

src/components/password-input/PasswordInput.tsx
```diff
@@ -65,36 +120,24 @@ export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
 
   const styledClassName = (feedbackType && styles[feedbackType]) || '';
 
-  const RevealOrHideIcon = showPassword ? HideIcon : RevealIcon;
-
   return (
     <div className={getClassNames(styles.root)}>
-      {label && <span className={styles.label}>{label}</span>}
+      <Label label={label} />
```
adelego, line 125: Weird usage for a label component. Did you do this? Why not `<Label>{label}</Label>`

src/components/text-input/TextInput.tsx
```diff
@@ -78,8 +78,19 @@ const InlineTextWrapper: React.FC<{
...
+  label,
+  fullWidth
+}) => {
+  return label ? (
+    <span
+      className={getClassNames(styles.label, {
+        [styles.fullWidth]: fullWidth
+      })}
+    >
+      {label}
```
adelego, line 91: See my comment above, but I think you should use the label children prop to define its text. 
A custom label should have the same props as a regular label. Can you change the Props type?

src/components/text-input/TextInput.module.scss
```diff
@@ -20,13 +20,18 @@
...
-  overflow-wrap: anywhere;
 
   color: colors.$dash-green-01;
   text-transform: uppercase;
   font-size: 10px;
   font-weight: 600;
   margin-bottom: 4px;
+  overflow-wrap: break-word;
+
+  &.fullWidth {
```
adelego, line 31: Why are you nesting this class? I'm not familiar with scss but it doesn't look necessary to me

src/components/password-input/PasswordInput.module.scss
pducolin: Sorry @oclero , removed my comment as I thought it was my PR 🤣 
I added a typgraphy label in my PR, you can wait for it and use it here

pducolin: It  is a new typo just for label, and it's also `uppercase`

src/components/text-input/TextInput.module.scss
pducolin: General sibling combinator https://developer.mozilla.org/en-US/docs/Web/CSS/General_sibling_combinator. Elements must have the same parent, and second one must appear after the first one, thought not necessarily immediately after

src/components/password-input/PasswordInput.tsx
oclero: I don't know all these React hooks

src/components/password-input/PasswordInput.tsx
oclero: To ensure it's typed as string?

src/components/text-input/TextInput.tsx
pducolin: I wouldn't mention the `inline`, custom should be fine

src/components/password-input/PasswordInput.module.scss
adelego: Thanks, @pducolin! In that case, I would suggest adding it to the typo file

src/components/password-input/PasswordInput.tsx
adelego: Why does it need to be a string? It could be an icon, a combination of text and span... It's up to the user to decide

sericaia, line 125: misses the `for` attribute, specially important for accessiblity

src/components/password-input/PasswordInput.tsx
```diff
@@ -65,36 +120,24 @@ export const PasswordInput = (props: PasswordInputProps): JSX.Element => {
...
+
       <div className={getClassNames(styles.container)}>
         <input
           ref={inputRef}
```
sericaia, line 129: needs an `id` to be linked with the label

src/components/text-input/TextInput.tsx
```diff
@@ -78,8 +78,19 @@ const InlineTextWrapper: React.FC<{
...
  * @param props
  */
-const Label: React.FC<{ label?: string }> = ({ label }) => {
-  return label ? <span className={styles.label}>{label}</span> : null;
+const Label: React.FC<{ label?: string; fullWidth?: boolean }> = ({
+  label,
+  fullWidth
+}) => {
+  return label ? (
+    <span
```
sericaia, line 86: use `<label>` instead (check https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label)

src/components/password-input/PasswordInput.tsx
adelego: In a functional component, by default everything is redefined at each render. useMemo is used for JSX element that you don't need to re-render everytime. Here the operation of creating the icon is not very costly but it's a good habbit to have. you can do something like:
```
const RevealOrHideIcon = useMemo(() => (showPassword ? HideIcon : RevealIcon), [showPassword])
```
This will only be called when `showPassword` changes

src/components/password-input/PasswordInput.tsx
oclero: We don't need the event in how it's handle in the code (this file). It's only an internal component, just for decoupling stuff. We don't need to provide such API.

src/components/password-input/PasswordInput.tsx
oclero: It could break the design or layout. But I'm going your way because it looks like it's always done like that.

src/components/password-input/PasswordInput.tsx
oclero: Nice! Thanks.

src/components/password-input/PasswordInput.module.scss
oclero: Perfect, I'll use it after rebasing

src/components/text-input/TextInput.module.scss
oclero: Improved in a new commit :)

src/components/password-input/PasswordInput.tsx
pducolin: maybe call the property `text` ? `label` in `Label` component sounds redundant

oclero: I've removed the prop and use children as Adèle says.

src/components/password-input/PasswordInput.module.scss
oclero: Actually, it's already in master.

goldins:
@ni0ki @maryiam @sericaia can this PR be updated or closed?

PR #136, opened 2020-06-26
Add stylelint

adelego:
This is a PR to include stylelint to the project. It consists of:
- adding stylelint, as well as plugins and configs to dependencies
- adding a config file
- add styelint to the pre-commit hook
- auto-fixing errors
- add better support for scss

### DIFF
--- a/src/typography/_typography.scss
+++ b/src/typography/_typography.scss
@@ -1,3 +1,5 @@
+/* stylelint-disable scale-unlimited/declaration-strict-value */
+
 /* Text */
 @mixin body-medium {
   font-family: 'Public Sans';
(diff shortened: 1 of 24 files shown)


src/typography/_typography.scss
```diff
@@ -1,3 +1,5 @@
+/* stylelint-disable scale-unlimited/declaration-strict-value */
```
pducolin, line 1: why is it disabled here?

adelego: This rule forces us to use variables for typos, among other properties. In this file, where we define typos, it doesn't make sense to enforce it

PR #137, opened 2020-06-29
refactor: remove value from slider label

pducolin:
Slight change to slider component, removing the actualized value from the label - can still be provided by the parent component.

### DIFF
--- a/src/components/slider/Slider.tsx
+++ b/src/components/slider/Slider.tsx
@@ -18,6 +18,7 @@ export interface SliderProps {
   step?: number;
   tooltip?: boolean;
   className?: string;
+  id?: string;
 }
 
 export const Slider = ({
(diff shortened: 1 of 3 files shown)


src/components/slider/Slider.tsx
```diff
@@ -18,6 +18,7 @@ export interface SliderProps {
   step?: number;
   tooltip?: boolean;
   className?: string;
+  id?: string;
```
oclero, line 21: Should I do the same for the TextInput and PasswordInput?

pducolin: If they are not acceptong any `HTMLInputElement` prop, I would say yes. I added this because if you use more than one slider on one page, they will have the same slider (`slider` by default)

PR #141, opened 2020-07-02
Feat/checkbox custom label

pducolin:
Add support for custom labels on checkbox.
Moved test to RTL.
Updated docs, handling state within the playground


src/components/checkbox/Checkbox.mdx
```diff
@@ -22,28 +22,33 @@ A checkbox component
...
-  <Checkbox disabled readOnly />
-  <Checkbox disabled checked readOnly />
-  <Checkbox checked label='Checked' />
-  <Checkbox label='Unchecked' />
+  {() => {
+      const [isChecked, setIsChecked] = React.useState(false);
+  const label = isChecked ? 'Checked' : 'Unchecked';
+return (
+  <>
+    <Checkbox label={label} checked={isChecked} onChange={e => setIsChecked(!isChecked)}/>
```
a-coruble, line 32: Even if we are in the documentation and everything should run smoothly, as you're depending on an older version in the state in the `onChange` handler, you should use the function form of the state update, like so :
`onChange={e => setIsChecked(prevIsChecked => !prevIsChecked)}`
For more information, [here's a quick summary](https://www.freecodecamp.org/news/get-pro-with-react-setstate-in-10-minutes-d38251d1c781/#state-updates-may-be-asynchronous) of why we should prefer that way when depending of the actual state to update the state

src/components/checkbox/Checkbox.tsx
```diff
@@ -48,11 +48,18 @@ export const Checkbox = ({
...
+        styles.checkboxContainer,
+        { [styles.disabled]: disabled },
+        className || ''
+      )}
+    >
       <input
         className={styles.hiddenCheckBox}
         type='checkbox'
         checked={checked}
+        disabled={disabled}
```
Domino9697, line 62: Good catch

src/components/checkbox/Checkbox.tsx
```diff
@@ -65,7 +65,7 @@ export const Checkbox = ({
       <StyledCheckbox checked={checked} disabled={disabled}>
         <Checkmark />
       </StyledCheckbox>
-      {label ? <span className={styles.text}>{label}</span> : null}
+      <span className={styles.label}>{label || ''}</span>
```
Domino9697, line 68: We put an empty string so the style of the label still applies here ?

pducolin: I can change this back. At some point while testing I used a label instead of a span, so I wanted the label to always be there.

src/components/checkbox/Checkbox.spec.tsx
```diff
@@ -1,36 +1,38 @@
 import * as React from 'react';
-import { mount } from 'enzyme';
-import 'jest-styled-components';
+import { render, screen } from '@testing-library/react';
 import { Checkbox } from './Checkbox';
+import userEvent from '@testing-library/user-event';
```
a-coruble, line 4: A little bit picky, but can you move this import on top of the relative `Checkbox` import ?

src/components/checkbox/Checkbox.tsx
```diff
@@ -48,17 +48,24 @@ export const Checkbox = ({
...
         type='checkbox'
         checked={checked}
+        disabled={disabled}
         {...otherProps}
       />
       <StyledCheckbox checked={checked} disabled={disabled}>
         <Checkmark />
       </StyledCheckbox>
-      {label ? <span className={styles.text}>{label}</span> : null}
+      <span className={styles.label}>{label || ''}</span>
```
a-coruble, line 68: Why have you changed this behaviour ? That means that if we don't have a label passed to the component, you're still going to render a useless label span inside the DOM, what do you think ?

src/components/checkbox/Checkbox.mdx
```diff
@@ -22,28 +22,33 @@ A checkbox component
...
+    <Checkbox disabled />
+    <Checkbox disabled checked />
+  </>
+  )}}
+
 </Playground>
 
-The example below shows an interactive `Checkbox`.
+`label` can be a React Component
```
a-coruble, line 40: I'm not sure that we need to precise it here as it's listed in the props as a ReactNode and having it here renders weirdly in the doc : 
image

src/components/checkbox/Checkbox.mdx
pducolin: will revert to `onChange={e => setIsChecked(e.currentTarget.checked)}`, ok?

src/components/checkbox/Checkbox.spec.tsx
pducolin: this should be done automatically IMO, looking for what is missing

src/components/checkbox/Checkbox.tsx
pducolin: I can change this back. At some point while testing I used a label instead of a span, so I wanted the label to always be there.

src/components/checkbox/Checkbox.mdx
pducolin: I added an header, now it looks like 
image

pducolin: I mostly did it to test it, as it's the main goal of my changes

src/components/checkbox/Checkbox.mdx
a-coruble: I would not recommend using the target DOM value, as we might not be in sync with the state of React for any reason. Prefer using the syntax I showed, it is a good practice according to [the React team too](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous)

src/components/checkbox/Checkbox.tsx
a-coruble: Yes, I would prefer avoiding useless DOM nodes in the render 👍

src/components/checkbox/Checkbox.mdx
a-coruble: Okay, cool, better with the header !! That works for me

src/components/checkbox/Checkbox.tsx
```diff
@@ -48,17 +48,24 @@ export const Checkbox = ({
...
         type='checkbox'
         checked={checked}
+        disabled={disabled}
         {...otherProps}
       />
       <StyledCheckbox checked={checked} disabled={disabled}>
         <Checkmark />
       </StyledCheckbox>
-      {label ? <span className={styles.text}>{label}</span> : null}
+      {label && <span className={styles.label}>{label}</span>}
```
a-coruble, line 68: Nice one 👍

PR #144, opened 2020-07-07
feat: add sharing center and exclamation icons

pducolin:
Adding sharing center and exclamation icons

image

image


src/design-tokens/icons/icons-defs.json
```diff
@@ -52,6 +53,7 @@
...
... 4.50183 16.9257C4.50166 16.9291 4.50153 16.9322 4.50141 16.935C4.50135 16.9363 4.5013 16.9376 4.50125 16.9388C4.50122 16.9394 4.50119 16.9403 4.50118 16.9406C4.50115 16.9414 4.50112 16.9422 6 17ZM12.0002 13C14.2955 13 15.3244 14.0312 15.8645 14.9876C16.1577 15.5068 16.3212 16.0408 16.4097 16.4544C16.4532 16.6581 16.4768 16.8242 16.4892 16.9331C16.4954 16.9873 16.4987 17.0265 16.5003 17.0483C16.5011 17.0592 16.5014 17.0656 16.5015 17.0673C16.5016 17.0682 16.5015 17.0678 16.5015 17.0662C16.5014 17.0654 16.5014 17.0643 16.5013 17.0629C16.5013 17.0622 16.5013 17.0614 16.5012 17.0606C16.5012 17.0601 16.5012 17.0594 16.5012 17.0592C16.5011 17.0585 16.5011 17.0577 18 17C19.4989 16.9423 19.4989 16.9415 19.4988 16.9407C19.4988 16.9404 19.4988 16.9395 19.4988 16.9389C19.4987 16.9377 19.4987 16.9365 19.4986 16.9351C19.4985 16.9323 19.4983 16.9292 19.4982 16.9258C19.4979 16.9189 19.4974 16.9107 19.4969 16.9011C19.4959 16.882 19.4943 16.8576 19.4922 16.8282C19.4878 16.7694 19.4809 16.6905 19.47 16.5942C19.4482 16.4023 19.41 16.1387 19.3433 15.8269C19.2112 15.2092 18.96 14.3682 18.4767 13.5124C17.4639 11.7188 15.4929 10 12.0002 10V13ZM18 15.5H6V18.5H18V15.5Z\" mask=\"url(#path-1-inside-1)\"/>",
```
lhoursquentin, line 56: According to the step by step guide we should remove all `fill` attributes, I see a `fill=\"white\"` here

pducolin: If I removed this fill, it breaks the design. As it is a mask, I think it should be kept. I will check what a mask is in svg, and update docs accordingly

lhoursquentin: All right thanks for the clarification, I just tried in red and it doesn't look like there is any white color in the way, so all good!

pducolin: Checked out on this. SVG's mask uses fill color luminance as the alpha (transparency) for the path using it. So here we have a white mask, and paths using it will have all pixel under the mask transparent.

.stylelintrc.json
```diff
@@ -1,5 +1,5 @@
 {
-  "ignoreFiles": ["./docs/*"],
+  "ignoreFiles": ["./docs/*", "**/*.tsx"],
```
ni0ki, line 2: This addition means that all styles inlined in .tsx files will be ignored ?
If so, I'm ok.

documentation/ICON_STEPBYSTEP.md
```diff
@@ -52,7 +52,7 @@ In order to avoid repetitive/manual work to create svg icons components, we choo
    import { MyIcon } from './MyIcon';
    ...
    <IconsWrapper>
-     ...
-     <MyIcon size={30} />
+   ...
+   <MyIcon size={30} />
```
lhoursquentin, line 56: Why remove the indentation here? The actual Icons.mdx file is indented there.

pducolin: This might be stylelint, will check it

.stylelintrc.json
pducolin: yes indeed, all files with `tsx` extension will be ignored by stylelint

documentation/ICON_STEPBYSTEP.md
pducolin: it's one of the linter changing this. Trying to fix the rule

lhoursquentin:
I don't see the usual docz update commit, is that expected?

pducolin:
> I don't see the usual docz update commit, is that expected?

probably not, it should be triggered automatically. I'm adding it manually.

PR #146, opened 2020-07-09
feat: infobox layout and icon adjustments

lhoursquentin:
- keep icon positioned at the top of the title (in case the title spans multiple lines)
- use a global padding for the infobox, rather than relying on the `<p>` element builtin margins which produced icon alignment issues
- allow passing an optional icon size, this avoids having to override the size from the css when using the element
- allow ReactNode type for the title, allowing to add some markup (links for instance) without needing an "as string" cast.
- avoid generating empty divs (for instance if description or buttons are not specified)
- fix a typo

BREAKING CHANGE: inner elements spacing and icon positioning adjustments


src/components/info-box/InfoBox.tsx
```diff
@@ -79,26 +78,29 @@ export const InfoBox = ({
...
       <div className={styles.headerRow}>
-        <div className={getClassNames(styles.icon, styles[size])}>
-          <InfoCircleIcon color={selectIconColor(severity)} />
+        <div className={styles.icon}>
+          <InfoCircleIcon size={iconSizePx} color={selectIconColor(severity)} />
         </div>
         <p className={getClassNames(styles.title, styles[size])}>{title}</p>
       </div>
-      {size === 'descriptive' ? (
+      {children !== undefined && size === 'descriptive' ? (
```
ni0ki, line 92: I think we could use `{children && [ ...] }` here so that we prevent it from being undefined and null.

lhoursquentin: Good point, but in that case passing a value equal to 0 as a children would prevent it from being displayed as well, for instance let's imagine displaying a number like `<InfoBox ..>{nb}</InfoBox>`, if `nb === 0` then nothing appears.

If we want to filter out `null` I'm guessing we should also filter `''`, empty array etc. What do you think of `(children === 0 || children) &&`?

PR #149, opened 2020-07-16
Feat/add password health icon

pducolin:
Adding password health icon


src/components/sidenav/Sidenav.tsx
```diff
@@ -7,6 +7,10 @@ export interface SidenavProps {
    * Enables thinner collapsed mode
    */
   collapsed?: boolean;
+  /**
+   * Side menu background color, default dash-green-00
+   */
+  backgroundColor?: string;
```
ni0ki, line 13: No other changes required to get this props working ?

pducolin: good catch, changed

PR #150, opened 2020-07-20
Feat/badge component

ni0ki:
Implements a basic badge.
No notion of status right now (such as success, error, etc...). This is all just props styling.

Preview of 2 different badges:

image


src/components/badge/Badge.module.scss
```diff
@@ -0,0 +1,11 @@
+.badge {
+  display: flex;
+  padding: 0 4px;
+  height: 20px;
+  font-size: 10px;
```
pducolin, line 5: isn't there a typography for it? Or could create one?

ni0ki: Good catch. I can use the label typography here :+1:

PR #151, opened 2020-07-20
Feat/add small alert

PILLOWPET:
* Implements two different sizes for the Alert component
* Fixes an issue with the icons' sizes which were not matching the designs

### Before

image

### Now

image

The icons on the left are a bit smaller (24px for medium and 20px for small) to match the designs.


src/components/alert/Alert.spec.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { mount, ReactWrapper } from 'enzyme';
 
 import { Alert, AlertProps } from '@components/alert/Alert';
-import { AlertSeverity } from '@components/alert/types';
+import { AlertSeverity, AlertSize } from '@components/alert/types';
```
pducolin, line 5: since you are refactoring, would you have time to move this test to RTL ?

src/components/alert/Alert.module.scss
```diff
@@ -29,10 +36,17 @@
   @include typography.body-small;
 
   color: colors.$white;
-  font-size: 18px;
   line-height: 20px;
+
+  &.medium {
+  font-size: 18px;
```
pducolin: isn't there a typography defined for this?

src/components/alert/Alert.tsx
```diff
@@ -26,26 +28,43 @@ export const Alert = ({
...
+      default:
+        return 24;
+    }
+  };
+
   const getIconBySeverity = (alertSeverity: AlertSeverity) => {
     switch (alertSeverity) {
       case AlertSeverity.ERROR:
-        return <CrossCircleIcon size={30} color='white' />;
+        return <CrossCircleIcon size={getIconSize()} color='white' />;
```
pducolin, line 45: cc @sericaia this might break your design in family dashboard

src/components/alert/Alert.module.scss
PILLOWPET: Unfortunately, there isn't

a-coruble:
Could you add some screenshots of the feature in the description of the PR ? This would help to see the differences

a-coruble:
Hhhm, just figured that your commits are unsigned, we will need you to re-sign your commits with your GPG key.

src/components/alert/Alert.module.scss
```diff
@@ -29,19 +36,28 @@
   @include typography.body-small;
```
ni0ki, line 36: If I follow your changes correctly, we are now either with a small or medium content.
Therefore, I think including this typography line is not relevant anymore (on top of that, we're redefining the line height to a different value right after), even misleading.

Could you define all the font property inside the `&.medium` scope, and use the `@mixin paragraph-small` in `&.small` ?
Thanks

ni0ki, line 36: If I follow your changes correctly, we are now either with a small or medium content.
Therefore, I think including this typography line is not relevant anymore (on top of that, we're redefining the line height to a different value right after), even misleading.

Could you define all the font property inside the &.medium scope, and use the @mixin paragraph-small in &.small ?
Thanks

src/components/alert/Alert.tsx
Domino9697: Yes about that ! We need to decide whether this is a breaking change :) 

I checked the designs and what Hadrien did follows them ! So IMO, this is a fix and not a breaking change !

src/components/alert/Alert.module.scss
```diff
@@ -26,12 +33,18 @@
...
 
 .description {
-  @include typography.body-small;
-
   color: colors.$white;
-  font-size: 18px;
-  line-height: 20px;
-  padding-left: 20px;
+
+  &.medium {
```
Domino9697, line 38: The paragraph medium typography does not match here ?

PILLOWPET: There is no `paragraph-medium` typography, only a `paragraph-small`

src/components/alert/Alert.spec.tsx
```diff
@@ -1,108 +1,57 @@
...
-      showIcon: true,
-      showCloseIcon: false,
-      children
-    });
-    expect(wrapper.exists('.root.success')).toEqual(true);
-    expect(wrapper.exists(CloseIcon)).toEqual(false);
+    const { container } = render(
+      <Alert showCloseIcon={false}>{children}</Alert>
+    );
+    expect(container).toMatchSnapshot();
```
Domino9697, line 46: While this is ok for now, I think we need to challenge the designers regarding this. The close icon button should have a tooltip text (title) on it.

src/components/alert/Alert.spec.tsx
```diff
@@ -1,108 +1,57 @@
...
-      children
-    });
-    expect(wrapper.exists('.description.small')).toEqual(true);
-    expect(wrapper.exists('.icon.small')).toEqual(true);
-    expect(wrapper.exists('.description.medium')).toEqual(false);
+    const { container } = render(
+      <Alert showCloseIcon={true}>{children}</Alert>
+    );
+    fireEvent.click(getByRole(container, 'button'));
+    expect(container).toMatchSnapshot();
```
Domino9697, line 55: Instead of having a snapshot here, we could test for the children text to appear before the click. Then test if it disappeared afterwards.

src/components/alert/Alert.module.scss
```diff
@@ -18,6 +17,14 @@
   background-color: colors.$accessible-validator-red;
 }
 
+.root.medium {
```
Domino9697, line 20: Could you move the medium and small classes with the root class ?

src/components/alert/Alert.module.scss
pducolin: we are adding typography as we need them, if the defined typography is `paragraph-medium` you can add it and use it

src/components/alert/Alert.tsx
pducolin: But did design changed in between?

Domino9697: Nope

src/components/alert/Alert.spec.tsx
```diff
@@ -1,108 +1,65 @@
...
-    );
-
   it('should create an alert with success, medium, and with icon closed if children is defined', () => {
     const children = 'Success test';
-    wrapper = createWrapper({
-      children
-    });
-    expect(wrapper.exists('.root.success')).toEqual(true);
-    expect(wrapper.exists('.root.medium')).toEqual(true);
+    const { container, queryByText } = render(<Alert>{children}</Alert>);
```
pducolin, line 9: for next iteration, you can skip using this thanks to `screen` https://testing-library.com/docs/dom-testing-library/api-queries#screen, you can find examples in the step by step guide to create a component here in docs

PR #152, opened 2020-07-20
feat: adding Warning theme for the button

MatMath:
For the InfoBox we need to have Warning theme button so the theme was added and link with the InfoBox
The default Danger theme is used for the Alert Primary button. 
The Alert theme and Subtle theme on blue need to be designed later.

InfoBoxButton
Button Warning


src/colors/colors.css
```diff
@@ -49,6 +49,9 @@
   --red-04: #f2ced6;
   --red-05: #f5d8de;
 
+  /* -- WARNING -- */
```
ni0ki, line 52: To be more consistent :
Is it possible to move this with the other "semantic" colors defined at the end of the file ? (validator_green, etc..)
Is it also possible to be more precise and name this color `--warning-red-00` ?

MatMath: I changed the naming, I thought having it closer to Red would have make sense but I no problem moving it lower.

PR #154, opened 2020-07-21
fix: textInput on safari and add disabled and readonly pseudo elements

sanatshah:
- Added offset margin to fix safari rendering issue
- Added selectors to inlineInput


sanatshah:
@a-coruble Hey not sure of your progress on Olivers PR https://github.com/Dashlane/ui-components/pull/135 , this is a short term solution to solve some problems on safari. Full solution would be to properly fix the readonly/disable css functionality as done in  his PR.  I can continue on his pr after this if that works =)

src/components/text-input/TextInput.module.scss
```diff
@@ -120,6 +122,7 @@
   display: inline-flex;
   height: 40px;
   margin: 0;
+  margin-left: -2px; /*offset spacing added by safari*/
```
ni0ki, line 125: Wouldn't this fix safari but breaks all other browsers ?

sanatshah: Won't break other browsers since the shifting is of the inline-text span box, not the full textInput box

PR #157, opened 2020-07-23
Feat/update dashlane business logo

akamkaro:
Update of Dashlane Business logo

image

image


lhoursquentin:
If we remove `Business` from the svg I'm guessing we should also remove it from the name of component, otherwise it might be a bit confusing.

But looking at our other "logo" components, it looks like this now is just a combination of [Logo](https://dashlane.github.io/ui-components/src-components-logo-logo) and [Dlogo](https://dashlane.github.io/ui-components/src-components-d-logo-d-logo), so I'm wondering if we should just remove this component altogether. It doesn't seem like it has any purpose left, you just need to use those two where needed:
dlogo
logo

akamkaro:
> If we remove `Business` from the svg I'm guessing we should also remove it from the name of component, otherwise it might be a bit confusing.
> 
> But looking at our other "logo" components, it looks like this now is just a combination of [Logo](https://dashlane.github.io/ui-components/src-components-logo-logo) and [Dlogo](https://dashlane.github.io/ui-components/src-components-d-logo-d-logo), so I'm wondering if we should just remove this component altogether. It doesn't seem like it has any purpose left, you just need to use those two where needed:
> dlogo
> logo
> 
> One other valid solution is to simply rename this component and modify it to use those two above.

Discussed with you and the designers, this logo should be totally independent from the other ones, as there are some differences (letter spacing...)

ni0ki:
For the notice, as a component is renamed, this should be a breaking change

Issue #76, opened 2020-04-24
Unwanted behavior when passing a container to a Tooltip


ni0ki:
Given the following code 
`() => {
  const containerRef = useRef(null)
  return (
    <div
      ref={containerRef}
      style={{
        border: '1px solid gray',
        width: '800px',
      }}
    >
      <Tooltip
        placement="right"
        title="This tooltip default position is right"
        type="success"
        container={containerRef}
      >
        <button>This tooltip default position is right</button>
      </Tooltip>
    </div>
  )
}`

I would expect the tooltip to be withing the container, as stated by the doc, not outside.
 
![image](https://user-images.githubusercontent.com/2590880/80197729-55cfa080-861f-11ea-94e6-ea70dc3029e6.png)

Maybe I'm using it the wrong way ?

PR #159, opened 2020-07-27
Fix/fix tooltip display

Domino9697:
If you look at the image below, you see that the tooltip is in the dom when it should not creating issues with height and width properties.

image


ni0ki:
Is it related to https://github.com/Dashlane/ui-components/issues/76 ? (Or does it has some impact on it ? )

Domino9697:
> Is it related to #76 ? (Or does it has some impact on it ? )

No I don't believe it is related ;) The issue I am fixing here is not related to the positioning of the tooltip.

PR #164, opened 2020-08-05
fix nav-link font-weight

Domino9697:


Domino9697:
@fmatias-dashlane I will do this next time :) This is a fix requested by the designers

PR #166, opened 2020-08-06
Feat/add button nav link

Domino9697:
The HorizontalNavMenu component only allows for links to be used. In our case, we also need buttons that the user can simply click with a callback.

I thus added a new component NavLinkButton.

image


src/components/horizontal-nav/horizontal-nav-button/HorizontalNavButton.module.scss
```diff
@@ -0,0 +1,6 @@
+.button {
```
ni0ki, line 1: I love that minimalist design :grinning:

Domino9697: Haha this is an override of default button styles actually :)

PR #168, opened 2020-08-11
feat: update accordion section title font width

sanatshah:
Update to font-weight for generic accordion section header in order to match design.


src/components/accordion/Accordion.module.scss
```diff
@@ -14,7 +14,7 @@
   flex-direction: row;
   font-size: 16px;
   font-weight: normal;
-  font-weight: 700;
+  font-weight: 600;
```
lhoursquentin, line 17: Can you remove the old duplicate `font-weight: normal` just above as well?

sanatshah: Oh man I don't know how i missed that 😮 😅

PR #169, opened 2020-08-12
feat: add isActive prop for tooltip

sanatshah:
Added ability to specify if the Tooltip is activated or not.


src/components/tooltip/Tooltip.tsx
```diff
@@ -28,6 +28,10 @@ export interface Props {
    * @default info
    */
   type?: 'info' | 'success' | 'error';
+  /**
+   * Whether or not the tooltip is activated. Defaults to true.
+   */
+  isActive?: boolean;
```
lhoursquentin, line 34: Personal preference, I'd rather have this as default == `false`, because `undefined` being the same as `true` is a bit less common, and thus would name it `disabled` or something similar. Non blocking of course.

src/components/tooltip/Tooltip.tsx
```diff
@@ -28,6 +28,10 @@ export interface Props {
    * @default info
    */
   type?: 'info' | 'success' | 'error';
+  /**
+   * Whether or not the tooltip is activated. Defaults to true.
```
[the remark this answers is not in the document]
sanatshah: I was going for "is the tooltip active" meaning will it be functioning or not. Maybe can rename it to isEnabled?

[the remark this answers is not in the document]
sanatshah: Oop just saw your comment below, Im ok with naming it disabled 🙂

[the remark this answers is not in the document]
lhoursquentin: Yeah then in the comment I'd just add something like: `tooltip will never show up (even when wrapped element is hovered) when disabled`  :slightly_smiling_face:

src/components/tooltip/Tooltip.tsx
```diff
@@ -215,6 +219,10 @@ export class Tooltip extends React.PureComponent<Props, State> {
       this.cloneChildren(title)
     );
 
+    if (this.props.isActive === false) {
```
ni0ki, line 222: Could this logic be set at the beginning of the render() function ? If `isActive` is false, I guess we don't need the above map operation which clones all the children ?

sanatshah: yes good idea, updated!

src/components/tooltip/Tooltip.tsx
sanatshah: Updated 🙂

PR #170, opened 2020-08-13
Tac 1735/button text multilined

bastienGranger:
Added the `white-space: nowrap;` property on button to avoid having multi-lined text.

**Before**
Screen Shot 2020-08-13 at 11 54 22 AM

**After**
Screen Shot 2020-08-13 at 4 02 51 PM


src/components/button/Button.module.scss
```diff
@@ -231,5 +231,6 @@
     font-weight: inherit;
     height: 100%;
     justify-content: center;
+    white-space: nowrap;
```
lhoursquentin, line 234: Looking at the screenshot what would happen if there's a third character? With `nowrap` it would overlap on the container padding right? I've disabled nowrap in a bunch of other places due to text overflowing out of the container (see screenshot below), I think forcing it is quite dangerous because you would need to check if it is displaying correctly in all the places this button is used and in all the languages.
button-overflow-nowrap

ni0ki: Agree that this is something we should pay attention about, but this seems more related to the layout chosen than to the button itself (or I may be missing something) ?

lhoursquentin: (In the screenshot I've shown the designer told me it was fine to wrap)

Yeah I'm not specifically against having `nowrap` by default, it's just that it's probably something we should have done from the start since now if we want to do this change we need to check all the places where this button is used, in all the languages.

ni0ki: Another remark, isn't the text wrapping because the button is not given enough space to grow ? If it's the case, shouldn't this be more a layout fix than a button one ?

bastienGranger: I maybe wrong but the layout should be fine and the button can grow. If you change the language to portuguese the button scale well

Screen Shot 2020-08-13 at 11 54 17 AM

bastienGranger: @lhoursquentin what would be the best solution?

lhoursquentin: I don't think there's an ideal solution unfortunately, either you modify your layout to give the button more space to grow or you add `nowrap` in the css where you are using the button.

bastienGranger:
Impact is to important if I do the change here. I opened a new MR in Leeloo

PR #175, opened 2020-08-19
Alert autodismiss and hook

fmatias-dashlane:
This PR adds three features for displaying alerts

- `useAlert` hook
  - new hook to allow programatically manipulating the alert
  - returns an object with
	- `alert` - the alert to be rendered
    - `show` - function to show the alert
    - `hide` - function to hide the alert
    - `setMessage` - function to change the displayed message
    - `setSeverity` - function to change the alert severity
- `dismissDelay` parameter
  - alert is automatically removed after a period of time
- `<AlertWrapper>`
  - export a wrapper to place the alerts at the bottom of its parent component
  - Note: not really customizable for lack of time, we can improve this along the way


src/components/alert/alert-wrapper/AlertWrapper.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import styles from './AlertWrapper.module.scss';
+
+export interface AlertWrapperProps {
+  children: React.ReactNode;
+}
+
+const hasContent = (children: React.ReactNode) => {
+  return React.Children.toArray(children).some(child => Boolean(child));
+};
```
a-coruble, line 11: Could you move this to the Utility folder ? I think that this could be reused in other places

src/components/alert/useAlert/useAlert.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import { Alert, AlertProps } from '@components/alert/Alert';
+
+const getRandomId = (): number => Math.floor(Math.random() * 1e6);
+
+export interface UseAlert {
+  alert: React.ReactChild | null;
```
a-coruble, line 7: You could use the `React.ReactNode` here as it also encapsulates the `null` value, string values and every other valid React element. `ReactChild` is missing some of the possible values, where ReactNode is more valide to represent the `children` that will be used.

src/components/alert/useAlert/useAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export type UseAlertParams = {
+  message: string;
+  dismissDelay?: number;
+} & Omit<AlertProps, 'children'>;
+
+export const useAlert = ({
+  message,
+  dismissDelay,
+  ...props
```
a-coruble, line 20: As this is not a component, I would rename this to `...params` to comply with your type name

src/components/alert/useAlert/useAlert.tsx
```diff
@@ -0,0 +1,50 @@
...
+}: UseAlertParams): UseAlert => {
+  const [id, setId] = React.useState<number | null>(null);
+
+  const show = React.useCallback(() => {
+    const newId = getRandomId();
+    setId(newId);
+  }, [setId]);
```
a-coruble, line 27: I think that your callback should also have the `getRandomId` in the dependencies array no ?

src/components/alert/useAlert/useAlert.tsx
fmatias-dashlane: True, no idea why I used `ReactChild` 🤦

sericaia (approved):

fmatias-dashlane:
@a-coruble @sericaia had to rebase with master, approvals got dismissed 😓 
One last approval and I'll get out of your hair 🙏

sericaia (approved):

a-coruble (approved):

PR #179, opened 2020-08-20
Feat/add bell active icon

PILLOWPET:
image


fmatias-dashlane:
I've seen some talks about a notification prop for icons, maybe @akamkaro can provide some feedback here.

a-coruble:
Indeed, there is a discussion ongoing concerning the topic of having a prop on the Icons to add a badge without having to create a new icon. Though, I don't know the requirements for this new icon you're proposing in terms of time for you. Also, we have a BellBadgeIcon which is similar to the one you're introducing here, so if it can, I would wait upon the new Icon component with a `badged` prop. Overall, we are discussing a refactor of how our icons are handled in this library.

PILLOWPET:
> I've seen some talks about a notification prop for icons, maybe @akamkaro can provide some feedback here.

Indeed, this would probably a better way to deal with the badge.

akamkaro:
Slack discussion about this: https://dashlane.slack.com/archives/CU3EKTDKP/p1597843971033500

PR #186, opened 2020-09-01
Fix/alert button and alert infoboxes

akamkaro:
- [x] Description text font-size changed
- [x] Icon size changed (default 20px (25 with margins))
- [x] Added Alert theme in button

**!!! Colors changes (difference between icon, title and description) won't be implemented in this PR !!!**

Figma:
- infoboxes: https://www.figma.com/file/I22yz2CF3Tnn2yr13snOsi/%5BKit%5D-%E2%80%93-Web?node-id=4728%3A277

image

image

image

image

image


ni0ki (approved):

src/components/info-box/InfoBox.module.scss
```diff
@@ -61,12 +61,16 @@
...
-  @include typography.body-small;
+  @include typography.paragraph-small;
   margin-top: 10px;
 
   a {
     color: inherit;
   }
+
+  &.alert {
+    color: colors.$functional-red-01;
```
MatMath, line 72: Why is the root:alert `red-00` but then force the body to be `red-01` ? 
All the other follow the same color for the title & body, any reason why this one different?

akamkaro: I follow the figma design : https://www.figma.com/file/I22yz2CF3Tnn2yr13snOsi/%5BKit%5D-%E2%80%93-Web?node-id=4728%3A277

src/components/button/Button.mdx
```diff
@@ -99,6 +100,21 @@ The default theme is `light` (white), but some buttons can be displayed in a `da
 
 > ⚠️ `danger` nature is not supported in `warning` theme.
 
+#### Alert
+
+<Playground className={{preview: 'playground-base'}} style={{ preview: {backgroundColor: red05} }}>
```
goldins, line 105: `className={{preview: 'playground-base'}}` does not look right here; `className` should be a string, right?

non-blocking: style-wise, this should ideally align with our current prettier formatting (namely spaces within braces) though I understand this does not get automatically formatted.

ni0ki: @goldins No, unfortunately DocZ does not allow easy customization of the <Playground> component. This was a little trick we made when we updated to the latest version, so that we can style it.
I agree however we would need to update that and write something cleaner

src/components/info-box/InfoBox.module.scss
akamkaro: @MatMath I removed it as i won't deal with this in this PR

src/components/info-box/InfoBox.tsx
```diff
@@ -83,6 +83,20 @@ const selectIconColor = (severity: InfoBoxSeverity) => {
...
+  if (iconSizePx !== undefined) {
+    return iconSizePx;
+  }
+
+  switch (size) {
+    case 'small':
+    case 'simple':
+      return 15;
+    case 'descriptive':
+      return 25;
```
akamkaro, line 96: I used those values because there is a margin/padding in the icon. 
So with 15, the visible part of the icon will be 12px (like in figma), and with 25 it will be 20 (like in figma)

a-coruble (approved):

ni0ki (approved):

ni0ki (approved):

PR #189, opened 2020-09-03
Feat/toast component

ni0ki:
Adds the Toast component

Based on the existing Alert, the Toast is an updated version from a design perspective, with support of CTA, positioning, and animations.

It can be used as a single component (static), or by a hook, which will provide all the animating stuff.

I choose not to remove the Alert right now as it is used, but rather mark it as deprecrated in the docs. We'll remove it once our apps has been updated.

(Sorry for the low quality flickering gif)
ui-components_toast


src/hooks/useToast/useToast.tsx
```diff
@@ -0,0 +1,114 @@
...
+    ToastWrapperTransition
+  >(ToastWrapperTransition.NONE);
+
+  const [toastMessage, setToastMessage] = React.useState<string>(message);
+  const [toastSeverity, setToastSeverity] = React.useState<
+    ToastSeverity | undefined
+  >(severity);
+
+  const show = React.useCallback(() => {
+    if (!id) {
```
lhoursquentin, line 46: Are you checking null here? 0 is a valid id right?

src/utility/warning-block/WarningBlock.module.scss
```diff
@@ -0,0 +1,19 @@
+@use '../../colors/colors.css';
+
+.warningBlock {
+  background-color: #fffdf6;
```
lhoursquentin, line 4: colors is included but unused

src/components/toast/Toast.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const { container, queryByText } = render(
+      <Toast showIcon={false} actionText={actionText} handleOnAction={() => {}}>
+        Testing Toast
+      </Toast>
+    );
+    expect(queryByText(actionText)).toBeInTheDocument();
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should react to CTA triggering', () => {
```
fmatias-dashlane, line 77: `should call handleOnAction when the CTA is clicked` 💅

src/components/toast/Toast.tsx
```diff
@@ -0,0 +1,138 @@
...
+  children,
+  severity = ToastSeverity.SUBTLE,
+  showIcon = true,
+  showCloseIcon = false,
+  size = ToastSize.MEDIUM,
+  actionText,
+  handleOnAction,
+  handleOnBeforeClose
+}: ToastProps) => {
+  const [close, setClose] = React.useState(false);
```
fmatias-dashlane, line 32: `isClosed` would be a more obvious name. `close` makes me think this is a function 😓

src/components/toast/Toast.tsx
```diff
@@ -0,0 +1,138 @@
...
+        return dashGreen00;
+      case ToastSeverity.STRONG:
+      case ToastSeverity.SUCCESS:
+        return white;
+      case ToastSeverity.ALERT:
+        return functionalRed01;
+    }
+  };
+
+  const getIconBySeverity = (toastSeverity: ToastSeverity) => {
```
fmatias-dashlane, line 64: We don't need the `toastSeverity` params in these functions, we already have access to the component `severity` prop.
Unless you want to remove these functions from inside the component, which I don't think is needed.

src/components/toast/Toast.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <div
+            className={getClassNames({
+              [styles.actionButtonOnly]: !showCloseIcon
+            })}
+          >
+            <ActionButton />
+          </div>
+        )}
+        {showCloseIcon && (
+          <Button
```
fmatias-dashlane, line 126: I don't really see a reason why `ActionButton` was refactored out and this close button wasn't 😅
Can you make it coherent? 🙏

src/components/toast/Toast.spec.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <Toast size={ToastSize.SMALL}>{children}</Toast>
+    );
+    expect(queryByText(children)).toBeInTheDocument();
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should not show close button if showCloseIcon is false', () => {
+    const children = 'Success test';
+    const { container, queryByText } = render(
+      <Toast showCloseIcon={false}>{children}</Toast>
```
fmatias-dashlane, line 50: This is the default case, we should also have a snapshot where `showCloseIcon={true}`.

src/components/toast/toast-wrapper/ToastWrapper.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export enum ToastWrapperTransition {
+  OPENING = 0,
+  CLOSING,
+  NONE
+}
+export interface ToastWrapperProps {
+  children: React.ReactNode;
+  onAnimationEnd: () => void;
+  transition: ToastWrapperTransition;
```
fmatias-dashlane, line 16: These two could be optional, in case someone wants to use this without the hook, right?

src/components/toast/Toast.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { dashGreen00, functionalRed01, white } from '@colors/colors';
+
+export interface ToastProps {
+  children: string;
+  severity?: ToastSeverity;
+  showIcon?: boolean;
+  showCloseIcon?: boolean;
+  size?: ToastSize;
+  actionText?: string;
+  handleOnBeforeClose?: () => boolean;
```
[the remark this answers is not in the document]
ni0ki:
```
  const handleOnClose = React.useCallback(() => {
    setClose(handleOnBeforeClose ? handleOnBeforeClose() : true);
  }, [handleOnBeforeClose, setClose]);
```

The base component, `Toast`, should close immediately when the user clicks on the 'X' button.
However, when using the hook, we don't want this to happen, but rather, we want to trigger the closing animation.

So `handleOnBeforeClose` returns a boolean, indicating to the `Toast` component if it can close by itself or not.
In the case of the hook, it will return `false`, because the hook will handle the removal of the toast.

It could also be used as a simple callback when the user closes the component.

src/hooks/useToast/useToast.tsx
fmatias-dashlane: Yes, this was a mistake I did previously, it should be `id !== null`. Good catch!

fmatias-dashlane (approved):
Good work @ni0ki 💪  Just some minor tweaks I identified, let me know when you have a final version to approve 🙏

src/components/toast/Toast.tsx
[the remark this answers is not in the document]
lhoursquentin: Ok got it thanks!

> It could also be used as a simple callback when the user closes the component.

If it's not too much trouble could you add an example using it in the mdx?

src/components/toast/toast-wrapper/ToastWrapper.tsx
ni0ki: Excellent catch. In my mind, we should try to avoid using the <ToastWrapper> manually, and I forgot to update the documentation about it.

However, there could be cases where it could make sense (like just wanting to position a Tooltip at the top / bottom), so let's keep it flexible :100:

src/hooks/useToast/useToast.tsx
ni0ki: Good catch. I guess you meant `id === null`, but indeed I need to protect against unwanted 0's

fmatias-dashlane: Yes, that! 🤦

sericaia:
🚨 I'm strongly against the creation of this component since it is mainly a copy of `Alert` (with animation). Can't we iterate on the Alert component? or have Alert with a `animate` prop?

PILLOWPET:
> 🚨 I'm strongly against the creation of this component since it is mainly a copy of `Alert` (with animation). Can't we iterate on the Alert component? or have Alert with a `animate` prop?

I think that ,in the end, the idea of this `Toast`component is to replace the `Alert`component. The name is different but we will progressively replace all instances of`Alert`component with `Toast` before removing it from `ui-components`.

ni0ki:
After talking with @sericaia, I rewrote this PR so that :

1. We get rid of the `Alert` component
2. We ensure the old `Alert` is seeing as renamed in Git so we keep its history

src/hooks/index.ts
```diff
@@ -1 +1 @@
-export { useAlert, UseAlert } from '@hooks/useAlert/useAlert';
+export { useToast, UseToast } from '@hooks/useToast/useToast';
```
sericaia: cool!

sericaia (approved):

PILLOWPET (dismissed):

PILLOWPET (dismissed):

ni0ki:
Added "Breaking change" as we are renaming the alert

PILLOWPET (dismissed):

src/components/toast/Toast.tsx
```diff
@@ -33,24 +44,72 @@ export const Toast = ({
...
+        onClick={handleOnClose}
+      >
+        <CloseIcon size={20} color={getIconColor()} />
+      </Button>
+    );
+  };
+
   const renderIcon = (toastSeverity: ToastSeverity, toastSize: ToastSize) => {
-    const icon = getIconBySeverity(toastSeverity);
+    const icon = getIconBySeverity();
```
fmatias-dashlane, line 112: You also don't need `toastSeverity` and `toastSize` here, you have them in the parent scope.

src/hooks/useToast/useToast.tsx
```diff
@@ -12,32 +22,53 @@ export interface UseToast {
...
   const [toastMessage, setToastMessage] = React.useState<string>(message);
   const [toastSeverity, setToastSeverity] = React.useState<
     ToastSeverity | undefined
   >(severity);
 
   const show = React.useCallback(() => {
-    const newId = getRandomId();
-    setId(newId);
-  }, [setId]);
+    if (id === null) {
```
fmatias-dashlane, line 52: I believe this will prevent new toasts from being displayed while another is already displayed, since the `id` isn't getting changed. Can you confirm?

We don't want that to happen because the user may perform multiple actions quickly that display toasts for the outcome.

src/components/toast/Toast.mdx
```diff
@@ -1,24 +1,24 @@
 ---
-name: Alert
+name: Toast
```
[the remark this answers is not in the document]
fmatias-dashlane: I saw after that you marked the following commit as the breaking change, even though the first commit is also one. Will you squash them together after?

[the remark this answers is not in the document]
ni0ki: I'll be careful to do so !

src/hooks/useToast/useToast.tsx
ni0ki: Indeed. This is a tricky one here, what do you suggest ?
Should we "stack" the Toast in this case (I guess it's the best case), or should we keep only one toast, but update it with latest content ?

fmatias-dashlane: My approach was to destroy the previously displayed toast and render a new one with the new content (destroying and creating new ones avoids automatic dismissals getting broken). This was achieved using the `key` prop.
I guess we may need to revisit this, in case we want stackable toasts but for the time being we can leave a note on the documentation regarding this behaviour and stick with it.

PR #190, opened 2020-09-04
Fix/icon color property update

ni0ki:
As stated on our internal Slack, descriptions for color-related parameters of the icon were not clear.
This super-tiny PR just aims to improve that.

Before:
image

After:
image


src/components/icons/types.ts
```diff
@@ -34,17 +34,17 @@ export interface Props {
...
   /**
    * The icon color when hovered
-   * @default dashGreen00 (dashlane default dark green)
+   * @default '#0E353D' (value of dashGreen00, dashlane dark green)
    */
   hoverColor?: string;
   /**
    * The icon color when active
-   * @default dashGreen00 (dashlane default dark green)
+   * @default '#0E353D' (value of dashGreen00, dashlane dark green)
```
akamkaro, line 47: Would it be possible to only accept values from the design system color palette?

ni0ki: It should be feasible if we use a custom type, but I'm not sure this is a good idea, as we might have edge cases where the dev needs a special color outside the palette ?

To be discussed I guess with people involved in the DS

akamkaro: In my opinion, it should be better to not have the possibility to use a custom color (or the color should be added to the palette). It is a way to control that the palette defined by designers is followed.
Do we have a confluence page where we could write those types of discussions?

akamkaro (approved):

a-coruble (approved):

PR #191, opened 2020-09-04
Feature/add new types to textinput

sericaia:
Adds support for email and number on the TextInput component


src/components/text-input/TextInput.tsx
```diff
@@ -2,11 +2,13 @@ import * as React from 'react';
 import { getClassNames } from '@utility/cssUtils';
 import styles from './TextInput.module.scss';
 
+type TextInputType = 'text' | 'email' | 'number';
```
lhoursquentin, line 5: Why not allow any input type?

lhoursquentin (approved):

src/components/text-input/TextInput.tsx
sericaia: password for instance is a different component

PR #192, opened 2020-09-05
chore: marginal improvements and dependency updates

goldins:
1) using prettier and lint-staged directly instead of pretty-quick.

pretty-quick use case according to the prettier docs: "Great for when you want an entire file formatting on your changed/staged files."

Using lint-staged directly instead: "Useful for when you want to use other code quality tools along with Prettier"

Also minimizing the number of libraries is nice.

2) updated pre-commit hooks and updated prettier/husky/lint-staged

Prettier doesn't need to run on all files; we can tweak this later if we need to.

Prettier should not run in parallel with eslint because eslint can catch errors before prettier fixes them.

`git add` no longer necessary as of a more recent version of husky or lint-staged (I forget the specifics)

3) [jest-watch-typeahead](https://github.com/jest-community/jest-watch-typeahead)

A neat little library that gives you a list of matching files/folders/tests when you're watching test

4) TS updated to 4.0.2


package.json
```diff
@@ -144,13 +144,12 @@
...
-    "*": [
-      "yarn pretty-quick"
+    "*.{tsx,ts,js,json}": [
+      "yarn prettier --write",
+      "eslint"
     ],
-    "*.{tsx,ts,js,json}": "eslint",
     "*.{tsx,ts,css}": [
-      "yarn stylelint --fix",
-      "git add"
```
sericaia, line 153: why are you removing `git add`?

sericaia: (I usually keep it in lint-staged - last operation)

goldins: https://github.com/okonet/lint-staged#reformatting-the-code

> **Prior to version 10, tasks had to manually include git add as the final step.** This behavior has been integrated into lint-staged itself in order to prevent race conditions with multiple tasks editing the same files. **If lint-staged detects git add in task configurations, it will show a warning in the console.** Please remove git add from your configuration after upgrading.

(emphasis mine)

I saw the warning and removed `git add` here

src/components/dropdown/__snapshots__/DropdownMenu.spec.tsx.snap
```diff
@@ -153,7 +153,7 @@ exports[`<DropdownMenu> Global render should have scroll enabled when using max
                           "attrs": Array [],
                           "componentStyle": ComponentStyle {
                             "componentId": "sc-bxivhb",
-                            "isStatic": true,
+                            "isStatic": false,
```
goldins, line 156: Note that this changed because `styled-components` was updated:

```diff
styled-components@^4.3.1:
- version "4.3.2"
+ version "4.4.1"
```

I can revert this and lock s-c if necessary.

https://github.com/styled-components/styled-components/issues/2796

> One of you likely has a newer version of s-c installed. We changed how `isStatic` is set in non-production environments. If you update to latest v4 and update your snapshots it should stay constant.

ni0ki (approved):

src/components/alert/__snapshots__/Alert.spec.tsx.snap
```diff
@@ -13,6 +13,7 @@ exports[`<Alert /> Should create a small alert 1`] = `
       >
         <svg
           class="svgWrapper animated"
+          style="--icon-width: 20px; --icon-height: 20px; --icon-idle-color: white; --icon-hover-color: white; --icon-active-color: white; --icon-rotate: none;"
           version="1.1"
           viewBox="0 0 20 20"
```
PILLOWPET, line 18: What caused the update to the snapshots? Is it the `jest`update?

PILLOWPET (approved):

lhoursquentin (approved):

src/components/alert/__snapshots__/Alert.spec.tsx.snap
goldins: Either jest or jsdom. From one of my many commit message:
> looks like jsdom/jest are now passing style properly. See `SvgWrapper`

I'm not sure if setting `--vars` as inline styles is what we should be doing, but that's the current behavior so these snapshots are more accurate now.

goldins:
I noticed I forgot to replace a .md instance of `pretty-quick` but I don't want to hold up the rest of these changes for that; I'll update that later.

PR #196, opened 2020-09-09
Feat/lockup update

ni0ki:
Update the Lockup component to latest Figma design. 

I need your opinion on the props, as I'm not totally fond of them so far.

- color : I find the name a bit misleading as I have to deal with fixed values here, as opposed to almost all other components.
- size : I first went for "small", "medium", etc... but Hugo highlighted that if we want to introduce an intermediate size, this will be difficult with that kind of naming. So I kept the height as possible values.

image


src/components/lockup/Lockup.tsx
```diff
@@ -1,41 +1,56 @@
...
-interface CustomProps {
-  /**
-   * Sets the logo for dark mode: "Dashlane" becomes white
-   * instead of dashGreen00.
-   * @default false
-   */
-  darkMode?: boolean;
+export interface LockupProps {
+  color: 'black' | 'dash-green' | 'mid-green' | 'pink' | 'white';
+  size: 32 | 39 | 48 | 64;
```
fmatias-dashlane, line 7: I would replace these possible values with enums that can grow indefinitely, this feels too much like magic 😅 

```tsx
export enum LockupColor {
  Black: 'black',
  Green: 'dash-green',
  MidGreen: 'mid-green',
  ...
}

export enum LockupSize {
  L1: 32,
  L2: 39,
  ...
}

export interface LockupProps {
  color: LockupColor,
  size: LockupSize
}

<Lockup size={LockupSize.L1} color={LockupColor.Green} />
```

What do you think?

src/components/lockup/Lockup.tsx
```diff
@@ -1,41 +1,56 @@
...
+      case 'black':
+        return black;
+      case 'dash-green':
+        return dashGreen00;
+      case 'mid-green':
+        return midGreen00;
+      case 'pink':
+        return pink00;
+      case 'white':
+        return white;
```
fmatias-dashlane, line 37: `getLockupSize` and `getLockupColor` could be replaced with an object outside the component, mapping the enum entries to the corresponding values. Not really a mandatory change though.
```ts
const colors = {
  [LockupColor.Green]: dashGreen00,
  [LockupColor.MidGreen]: midGreen00
}
const sizes = {
  [LockupSize.L1]: { width: 121.5, height: 32 },
  ...
}

export const Lockup = ({ color, size }: LockupProps) => {
  const colorValue = colors[color];
  const sizeValue = sizes[size];
  ...
}
```

fmatias-dashlane (approved):
Good work! Left you some improvement suggestions but it looks good 👌

src/components/lockup/Lockup.tsx
ni0ki: Could be a good idea for the color.
About the size, I'm not sure we'll benefit from this, as if designers introduce a new intermediate size, we'll need to shift the whole enum values :| (or have something like L1.5).

akamkaro: I agree with @ni0ki . Replacing sizes in px with L1, L2... would be problematic if designers introduce new values.

akamkaro (dismissed):

src/components/lockup/Lockup.tsx
fmatias-dashlane: Yes, that's a very valid point 😅

src/components/lockup/Lockup.tsx
```diff
@@ -2,48 +2,46 @@ import * as React from 'react';
 import { SvgWrapper } from '@components/icons/utils/SvgWrapper';
 import { black, dashGreen00, midGreen00, pink00, white } from '@colors/colors';
 
+export enum LockupColor {
+  Black = 'black',
+  DashGreen = 'dash-green',
+  MidGreen = 'mid-green',
+  Pink = 'pink',
```
akamkaro, line 9: Do you want to keep this one?

src/components/lockup/Lockup.mdx
```diff
@@ -16,13 +16,16 @@ Dashlane Lockup component
 
 <Playground
   style={{
-    display: 'flex',
-    flexDirection: 'row',
-    justifyContent: 'space-around'
+    preview: {
```
fmatias-dashlane, line 19: Both commits are marked as breaking changes, maybe you could squash them together and have just one breaking change? 🙏

src/components/lockup/Lockup.tsx
```diff
@@ -2,48 +2,46 @@ import * as React from 'react';
...
+  Pink,
+  White
+}
+
+export enum LockupSize {
+  Size32,
+  Size39,
+  Size48,
+  Size64
+}
```
fmatias-dashlane: Much better than my suggestion, good one!

akamkaro (approved):

src/components/lockup/Lockup.mdx
ni0ki: Is it ok for you if I squash them when merging ?
Rebasing is a pain on ui-components because the doc is regenerated each time :|

fmatias-dashlane (approved):

